PR #5, opened 2024-12-09
Skip moxie reward for paid scouts

Devorein:


apps/admin/app/api/partners/moxie/route.ts
```diff
@@ -104,6 +132,21 @@ export async function GET() {
         }))
       })
     });
+    await Promise.all(
+      Object.values(scoutMoxieAmounts).map(async ({ userId, amount }) => {
+        await prisma.partnerRewardEvent.create({
```
mattcasey, line 137: please save this immediately after each user, in case it fails halfway

mattcasey (changes requested):
see comment. also, in case it does fail, i would sort the users by something like created date, or amount of reward so that we will be able to know where to continue it once we resolve the issue

PR #11, opened 2024-12-11
Common Scout page

valentinludu:


packages/scoutgame-ui/src/components/scout/ScoutPage.tsx
```diff
@@ -26,17 +33,14 @@ export const scoutTabMobileOptions: TabItem[] = [
   { label: 'New Scouts', value: 'new-scouts' }
 ];
 
-export function ScoutPage({
+export async function ScoutPage({
```
mattcasey, line 36: what is the purpose of making this async/server-only?

packages/scoutgame/src/utils/async.ts
```diff
@@ -45,3 +45,9 @@ export async function safeAwaitSSRData<T>(promise: Promise<T>) {
     }
   });
 }
+
+export async function delay(ms: number = 1) {
```
mattcasey, line 49: nice util :)

packages/scoutgame/src/builders/aggregateTokensPurchased.ts
```diff
@@ -0,0 +1,12 @@
+import { prisma } from '@charmverse/core/prisma-client';
```
mattcasey, line 1: thanks for adding this to a lib. NBD but i think it would make more sense inside a scouts folder

mattcasey (approved):

packages/scoutgame-ui/src/components/scout/ScoutPage.tsx
valentinludu: We have child components that are async/server-only

PR #13, opened 2024-12-11
Merging scoutgame accounts

Devorein:


packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+  if (primaryUser.builderStatus !== null && secondaryUser.builderStatus !== null) {
+    throw new Error('Can not merge two builder accounts');
+  }
+
+  // The id of the user to retain
+  const retainedUserId = secondaryUser.builderStatus !== null ? secondaryUser.id : userId;
+
+  // The id of the user to merge into the retained user
+  // The merged account must not be a builder
+  const mergedUserId = secondaryUser.builderStatus === null ? secondaryUser.id : userId;
```
mattcasey, line 48: please put this logic in the frontend, so that this function is more deterministic. Just take a primaryId and secondaryId (or farcaster/telegram of the secondary account). The caller/client should know which userId is being kept, and this function shouldn't just change that fact based on the db state. it would be better to just throw an error if the input is invalid

mattcasey: the frontend also needs to know if the user has to be logged out or not (because the button will say "Merge and Log out") so there shouldnt be a reason to put the logic this far down

packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+        farcasterName: farcasterId ? mergedUser.farcasterName : undefined,
+        telegramId
+      };
+
+      // Only merge the new profile if none of the users are builders
+      if (selectedProfile === 'new' && mergedUser.builderStatus === null && primaryUser.builderStatus === null) {
+        updatedUserData.avatar = mergedUser.avatar;
+        updatedUserData.displayName = mergedUser.displayName;
+        updatedUserData.bio = mergedUser.bio;
+        updatedUserData.email = mergedUser.email;
```
mattcasey, line 79: i don't think we should override email. at least check if it exists first

packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+        updatedUserData.walletENS = mergedUser.walletENS;
```
mattcasey, line 80: this should not be overridden unles we're changing your wallets

packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+      }
+
+      // Detach the identities from the merged user
+      await tx.scout.update({
+        where: {
+          id: mergedUserId
+        },
+        data: {
+          email: null,
+          farcasterId: farcasterId ? null : undefined,
```
mattcasey, line 90: these can all be safely set to null no matter what since you're pulling them over to the primary user

packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+      email: true
+    }
+  });
+
+  await prisma.$transaction(
+    async (tx) => {
+      const updatedUserData: Prisma.ScoutUpdateInput = {
+        farcasterId,
+        farcasterName: farcasterId ? mergedUser.farcasterName : undefined,
+        telegramId
```
mattcasey, line 71: This method should safeguard in case telegramId is passed in as "null", where it could remove the existing telegramId if one exists. I don't think it does at tthe moment

packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+      };
+
+      // Only merge the new profile if none of the users are builders
+      if (selectedProfile === 'new' && mergedUser.builderStatus === null && primaryUser.builderStatus === null) {
```
mattcasey, line 75: instead of checking for `mergedUser.builderStatus === null && primaryUser.builderStatus === null` here, i think this method should simply throw an error if it was given invalid input for selectedProfile

packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+      select: {
+        builderStatus: true,
+        id: true
+      }
+    })
+  ]);
+
+  if (primaryUser.builderStatus !== null && secondaryUser.builderStatus !== null) {
+    throw new Error('Can not merge two builder accounts');
+  }
```
mattcasey, line 41: you should check if the secondaryUser.deletedAt is set, and maybe just throw an error. this avoids someone calling it twice, but the 2nd time the secondary user would not have the correct attributes anymore

packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+        },
+        data: {
+          scoutId: retainedUserId
+        }
+      });
+
+      await tx.scout.update({
+        where: { id: mergedUserId },
+        data: {
+          deletedAt: new Date()
```
mattcasey, line 113: why not do this around line 84 with the other updates?

packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+        }
+      });
+
+      await tx.scoutMergeEvent.create({
+        data: {
+          mergedFromId: mergedUserId,
+          mergedToId: retainedUserId,
+          mergedRecords: {
+            farcasterId,
+            telegramId
```
mattcasey, line 123: i would include email here too

packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+      });
+
+      await tx.scoutGameActivity.updateMany({
+        where: {
+          userId: mergedUserId
+        },
+        data: {
+          userId: retainedUserId
+        }
+      });
```
mattcasey, line 207: also builderEvent probably? Just in case.. i think we are using it now for non-builder things sometimes.

mattcasey: Also: ScoutSocialQuest, ScoutDailyClaimEvent, and ScoutDailyClaimStreakEvent

packages/scoutgame/src/users/mergeUserAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+    async (tx) => {
+      const updatedUserData: Prisma.ScoutUpdateInput = {
+        farcasterId,
+        farcasterName: farcasterId ? mergedUser.farcasterName : undefined,
+        telegramId
+      };
+
+      // Only merge the new profile if none of the users are builders
+      if (selectedProfile === 'new' && mergedUser.builderStatus === null && primaryUser.builderStatus === null) {
+        updatedUserData.avatar = mergedUser.avatar;
```
mattcasey, line 76: also pull over 'path'

packages/scoutgame/src/users/mergeUserAccount.ts
Devorein: good point. updated

packages/scoutgame/src/users/mergeUserAccount.ts
Devorein: we are changing the wallet though. All the wallets of the previous holder will wallets of the new holder but just like email I will not overwrite it if it already exists.

packages/scoutgame/src/users/mergeUserAccount.ts
Devorein: agree. Updated.

packages/scoutgame/src/users/mergeUserAccount.ts
Devorein: updated. selected profile of null must be provided if any of the accounts is a builder otherwise we will throw an error

packages/scoutgame/src/users/mergeUserAccount.ts
Devorein: If we return success it will fire off another function to reconcile the stats. Best would be to throw an error as this is not going to be possible from the frontend

packages/scoutgame/src/users/mergeUserAccount.ts
Devorein: added wallets too as after merge it won't be possible to know which one belonged to which user

packages/scoutgame/src/users/mergeUserAccount.ts
Devorein: builder event is not necessary as we will always keep the builder intact, and there are unique validation on the rest of the records to not add it for those. eg partner rewards, scout social etc.

packages/scoutgame/src/users/mergeUserAccount.ts
Devorein: the client does know which user will be kept based on the builder status. the reason primaryId or secondaryId is not passed to the action is because it can be tampered with easily. we already have the primary user which is the logged in user and the secondary user will be obtained from the auth data provided from tg or farcaster.

mattcasey (approved):

PR #17, opened 2024-12-12
Show the percent reward when buying cards

mattcasey:


apps/scoutgame/app/api/builders/search/route.ts
```diff
@@ -1,20 +1 @@
...
-      search
-    });
-
-    return Response.json(result);
-  } catch (error) {
-    log.error('Error requesting user from farcaster', { error, search });
-    return new Response(`Unknown error: ${(error as Error).message}`, { status: 500 });
-  }
-}
+export * from '@packages/scoutgame-router/api/builders/search/route';
```
mattcasey, line 1: a new pattern to reuse api endpoints and other app endpoints. unfortunately, symlinks are not respected by next.js when i test in dev

jest.teardown.ts
```diff
@@ -0,0 +1,25 @@
+import { log } from '@charmverse/core/log';
```
mattcasey, line 1: i added this because my VS Code expected it to be here when it ran jest. maybe we could just use the teardown from scoutgame app but i didnt want to get bogged down in jest config

packages/scoutgame/src/points/dividePointsBetweenBuilderAndScouts.ts
```diff
@@ -7,8 +7,8 @@ import { builderPointsShare, scoutPointsShare } from '../builderNfts/constants';
 import { calculateEarnableScoutPointsForRank } from '../points/calculatePoints';
 
 const nftTypeMultipliers: Record<BuilderNftType, number> = {
-  starter_pack: 1,
-  default: 10
+  starter_pack: 0.1,
```
mattcasey, line 10: i find it easier to think of starter pack's value relative to 'default', versus the other way around. for example if we added a 3rd NFT, it makes more sense if they are all relative to default and not starter_pack

Devorein (approved):

PR #27, opened 2024-12-16
Rank new scouts based on predicted points for this week

mattcasey:
This is more accurate. Doesn't seem to be much slower performance-wise.

I will fix tests once CI lets me know :)

image


packages/scoutgame/src/points/dividePointsBetweenBuilderAndScouts.ts
```diff
@@ -43,50 +43,29 @@ export async function dividePointsBetweenBuilderAndScouts({
     throw new InvalidInputError('Invalid rank provided. Must be a number greater than 0');
   }
 
-  const nftPurchaseEvents = await prisma.nFTPurchaseEvent.findMany({
-    where: {
-      builderEvent: {
-        week: {
-          lte: week
+  // Calculate the total number of NFTs purchased by each scout
```
mattcasey, line 46: by removing the NFT history query here, we can be much faster if we just retrieve nft purchases for all builders ta once

packages/scoutgame/src/points/getWeeklyPointsPoolAndBuilders.ts
```diff
@@ -1,20 +1,70 @@
...
-
-  const { normalisationFactor, totalPoints } = await getPointsCountForWeekWithNormalisation({ week });
+export type PartialNftPurchaseEvent = {
+  scoutId: string;
+  tokensPurchased: number;
+  builderNft: { nftType: BuilderNftType; builderId: string };
+};
 
-  const weeklyAllocatedPoints = await getCurrentWeekPointsAllocation({ week });
+export async function getWeeklyPointsPoolAndBuilders({
```
mattcasey, line 17: Add nft purchases here and also make the queries run in parallel

Devorein (approved):

PR #29, opened 2024-12-17
Fix scout ranks

valentinludu:


packages/scoutgame/src/scouts/getScouts.ts
```diff
@@ -62,6 +62,9 @@ export async function getScouts({
         }
       },
       take: limit,
+      orderBy: {
```
Devorein, line 65: we need to do the ascending or descending order based on payload. There is an example below on line 113

Devorein (changes requested):

packages/scoutgame/src/scouts/getScouts.ts
valentinludu: We want that, but only for the first 200. That means we need to take the first 200 from 1 to 200 and then sort it asc desc.

Devorein (approved):

PR #32, opened 2024-12-17
Scout Carousels Redesign

valentinludu:


apps/telegram/app/(general)/u/[path]/not-found.tsx
```diff
@@ -1,4 +1,4 @@
-import { NotFoundPage } from 'components/layout/NotFoundPage';
+import { NotFoundPage } from '@packages/scoutgame-ui/components/common/NotFoundPage';
```
Comment (mattcasey) [inline apps/telegram/app/(general)/u/[path]/not-found.tsx:1]: thanks for moving

packages/scoutgame-ui/src/components/scout/ScoutPage.tsx
```diff
@@ -71,8 +73,8 @@ export async function ScoutPage({
             gap: 2
           }}
         >
-          <Suspense fallback={<LoadingCards count={3} />}>
-            <ScoutPageCarousel />
+          <Suspense key={carousel} fallback={<LoadingCards count={3} />}>
```
mattcasey, line 76: we shouldn't need to re-query the data every time you change tabs. ideally it just loads once on page load.

mattcasey (approved):

packages/scoutgame-ui/src/components/scout/ScoutPage.tsx
valentinludu: Yep. I updated this and the async function removal so it doesn't load again

PR #39, opened 2024-12-18
dont include daily quest in points earned as builder

mattcasey:


packages/scoutgame/src/points/getPointStatsFromHistory.ts
```diff
@@ -59,37 +59,31 @@ export async function getPointStatsFromHistory({
...
-              builderId: {
-                not: userId
-              },
-              type: 'gems_payout'
-            }
+        event: {
+          type: 'gems_payout'
+        },
+        // we need to include points received when you own your own NFT
+        activities: {
```
mattcasey, line 82: luckily we have this activities collection! I am not sure how else to capture this information, other than an optional field on points receipt

Devorein (approved):

PR #41, opened 2024-12-19
remove deleteMany in test setup

mattcasey:
clearing a collection results in flaky tests. tests can just use custom inputs to keep their data separate


packages/scoutgame/src/builderNfts/getCurrentWeekPointsAllocation.ts
```diff
@@ -4,7 +4,7 @@ import { prisma } from '@charmverse/core/prisma-client';
 export const pointsPerActiveBuilder = 2_500;
 
 export async function getCurrentWeekPointsAllocation({ week }: { week: string }) {
-  if (typeof week !== 'string' || !week.match(/^\d{4}-W\d{2}$/)) {
+  if (typeof week !== 'string') {
```
mattcasey, line 7: this method is called internally and is unecssary

PR #50, opened 2024-12-24
Fixed telegram account connect QR code

Devorein:


packages/scoutgame/src/telegram/verifyTelegramTokenAction.ts
```diff
@@ -0,0 +1,92 @@
...
+import type { TelegramClient } from 'telegram';
+import { Api } from 'telegram';
+import { UpdateConnectionState } from 'telegram/network';
+
+import { encrypt } from '../utils/crypto';
+
+const TIMEOUT_MS = 30000;
+
+// eslint-disable-next-line
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
```
valentinludu, line 15: We have a delay func in packages/utils/async

packages/scoutgame-ui/src/components/accounts/components/TelegramConnect.tsx
```diff
@@ -66,67 +51,128 @@ export function TelegramConnect({ user }: { user: UserWithAccountsDetails }) {
...
+  });
+
+  const { executeAsync: removeTelegramClient } = useAction(removeTelegramClientAction);
+
+  const { executeAsync: generateTelegramQrCode, isExecuting: isGeneratingQrCode } = useAction(
+    generateTelegramQrCodeAction,
+    {
+      onSuccess: (data) => {
+        if (data.data?.qrCodeImage) {
+          setQrCode(data.data.qrCodeImage);
```
valentinludu, line 80: I don't think you need a state for qrCode since you can use the result?.data?.qrCodeImage.
And when you verifyTelegramToken() -> onSuccess you can use `reset` to reset the qrCodeImage.

valentinludu: Nvm. I see that this func needs to be after the above one since it uses `verifyTelegramToken`. That means you cant `reset` qrCodeImage from above.

valentinludu (approved):
Looks good.

PR #57, opened 2024-12-27
Update referral scoring for nft purchase by referee

Devorein:


mattcasey (approved):

packages/scoutgame/src/builderNfts/recordNftMint.ts
```diff
@@ -213,6 +214,83 @@ export async function recordNftMint(
     log.error('Error completing quest', { error, builderId: builderNft.builderId, questType: 'scout-starter-card' });
   }
 
+  try {
+    // Find the referrer of the scout first
+    const referrer = await prisma.scout.findFirst({
```
mattcasey, line 219: this should live outside in a separate funciton

packages/scoutgame/src/builderNfts/recordNftMint.ts
```diff
@@ -213,6 +214,83 @@ export async function recordNftMint(
...
+          }
+        }
+      },
+      select: {
+        id: true
+      }
+    });
+
+    if (referrer) {
+      const referralCodeEvent = await prisma.referralCodeEvent.findFirst({
```
mattcasey, line 235: We do'nt have a database guarantee that refereeId is unique; you should only do one query. i dont think u need to query referrer first. just query the referallcodeevent

mattcasey (changes requested):
see comments

packages/scoutgame/src/builderNfts/recordNftMint.ts
```diff
@@ -213,6 +214,83 @@ export async function recordNftMint(
...
+        await prisma.$transaction(async (tx) => {
+          await tx.builderEvent.create({
+            data: {
+              type: 'referral_bonus',
+              season: currentSeason,
+              description: 'Received points for a referred user scouting a builder',
+              week: getCurrentWeek(),
+              builder: {
+                connect: {
+                  id: referrer.id
```
mattcasey, line 255: at the moment there's no guarantee that referrer.id is the same as referralCodeEvent.builderEvent.builderId

packages/scoutgame/src/builderNfts/recordNftMint.ts
```diff
@@ -213,6 +214,83 @@ export async function recordNftMint(
...
+        where: {
+          refereeId: scoutId
+        },
+        select: {
+          id: true
+        }
+      });
+
+      // If the referral code event exists and doesn't have a bonus event, create one
+      if (referralCodeEvent && !referralCodeEvent.bonusEvent) {
```
mattcasey, line 245: where is this bonusEvent added so that it doesnt get triggered again?

PR #71, opened 2024-12-31
Improve scout carousel tabs performance

valentinludu:
### Why
Using the query param as a state manager makes the app slow. 

### What
Using a react state makes the change of the tabs instant.


mattcasey (approved):
Looks good, and I prefer the state not bbeing added to the URL anyway

I'm assuming this works because the state is only triggering redraws inside the carousel component. I wonder if we could be doing better with memoization + props in the higher-level components

PR #88, opened 2025-01-03
Builder partner pages

Devorein:
[Card](https://app.charmverse.io/cvt-scout-game/build-partner-landing-pages-7087070548492638)


valentinludu (commented):
These pages should be in `telegram` app folder too.

You can use the approach of packages/scoutgame-router.

We are using it for API's and it will work with pages too.

packages/scoutgame-ui/src/components/partners/PartnerLandingPage.tsx
```diff
@@ -0,0 +1,383 @@
...
+      {additionalContent}
+    </ContainerStack>
+  );
+}
+
+function HeroSection({ heroSubtitle, heroImage }: { heroSubtitle: ReactNode; heroImage: string }) {
+  return (
+    <Stack sx={{ position: 'relative' }}>
+      <Hidden mdDown>
+        <Image
```
valentinludu, line 95: This is why sometimes you need zIndex on elements since you have an absolute element on the screen.
Better to add it as css background props in any div you want

```
      sx={{
        backgroundImage: {
          xs: undefined,
          md: 'url(/images/home/landing-bg.png)'
        },
        backgroundSize: 'cover',
        backgroundPosition: 'center',
        backgroundRepeat: 'no-repeat'
      }}
```

valentinludu (approved):
Looks good.

Left some comments. 

Making one Component and just sharing it through all the pages would be an improvement on how we create pages.

packages/scoutgame-ui/src/components/partners/PartnerLandingPage.tsx
Devorein: thnx updated and its a lot better + easier

PR #106, opened 2025-01-06
Update mixpanel scripts

valentinludu:
Created scripts for:
- add sign_up events for list of users
- delete user profiles


scripts/migrations/2025_01_06_addMixpanelSignupEvents.ts
```diff
@@ -0,0 +1,57 @@
...
+import { uuidFromNumber } from '@packages/utils/uuid';
+import fs from 'node:fs';
+
+async function trackMixpanelSignupEvents(userIds: string[]) {
+  const scouts = await prisma.scout.findMany({
+    where: {
+      id: {
+        in: userIds
+      },
+      onboardedAt: {
```
mattcasey, line 13: i think we should stick with our logic, which means that sign_up events occur before you are onboarded and being onboarded is not required

packages/mixpanel/src/updateUserProfile.ts
```diff
@@ -36,3 +37,38 @@ export function batchUpdateMixpanelUserProfiles(users: { userId: string; profile
     }))
   );
 }
+
+export async function deleteMixpanelProfiles(users: { id: string }[]) {
```
mattcasey, line 41: can you put this in its own folder

mattcasey (approved):

packages/mixpanel/src/updateUserProfile.ts
valentinludu: done

scripts/migrations/2025_01_06_addMixpanelSignupEvents.ts
valentinludu: done

PR #110, opened 2025-01-06
Add season to social quest

Devorein:


mattcasey (approved):

mattcasey (approved):
looks OK but don't we still need to account for quests that are not resettable?

PR #131, opened 2025-01-07
manage subscriptions when user edits email settings

mattcasey:
a lot of this is just copied from charmverse


packages/users/src/updateUserEmailSettings.ts
```diff
@@ -1,23 +1,62 @@
...
+
+  if (original.email !== updatedUser.email || original.sendMarketing !== updatedUser.sendMarketing) {
+    try {
+      if (!updatedUser.email) {
+        // remove from Loops and Beehiiv
+        await deleteLoopsContact({ email: original.email! });
+        await deleteBeehiivSubscription({ email: original.email! });
+      } else {
+        await registerLoops({ ...updatedUser, oldEmail: original.email }, getPlatform());
+        await registerBeehiiv({ ...updatedUser, oldEmail: original.email });
```
Devorein, line 54: This looks a bit weird. When will !updatedUser.email ever be truthy, email is always required. And shouldn't we delete the beehive and loops data when they toggle off the send marketing flag?

Devorein (changes requested):
left one comment regarding when to unsubscribe the user otherwise looks good

packages/users/src/updateUserEmailSettings.ts
mattcasey: @Devorein you're right, email is required. i'll just remove that i guess. it wsa copied from charmverse where we allowed no email

the register methods handle sendMarketing = false already. insetad of deleting they just unsubscribe the user

PR #132, opened 2025-01-07
Delete beehiv loops

mattcasey:
When removing deleted users we should remove them from loops and beehiiv as well, so I repurposed the mixpanel task to be a job for any external profiles.

This PR also makes it so we only look back 2 hours. The job runs once an hour so that should be plenty of time to check.

### DIFF
--- a/apps/cron/src/tasks/syncExternalUserProfiles/syncExternalUserProfilesTask.ts
+++ b/apps/cron/src/tasks/syncExternalUserProfiles/syncExternalUserProfilesTask.ts
@@ -8,6 +8,8 @@ import { deleteExternalProfiles } from './deleteExternalProfiles';
 
 const perBatch = 1000;
 
+const deletedAtLoobackHours = 2; // job runs once an hour, 2 hours gives a little overlap
+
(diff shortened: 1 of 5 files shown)


apps/cron/src/tasks/syncExternalUserProfiles/syncExternalUserProfilesTask.ts
```diff
@@ -1,11 +1,15 @@
...
-import { deleteMixpanelProfiles } from '@packages/mixpanel/deleteUserProfiles';
 import { batchUpdateMixpanelUserProfiles } from '@packages/mixpanel/updateUserProfile';
 import type { MixPanelUserProfile } from '@packages/mixpanel/updateUserProfile';
+import { DateTime } from 'luxon';
+
+import { deleteExternalProfiles } from './deleteExternalProfiles';
 
 const perBatch = 1000;
 
+const deletedAtLoobackHours = 2; // job runs once an hour, 2 hours gives a little overlap
```
valentinludu, line 11: I think there is a typo in the const name. `LookBack` maybe?

valentinludu (approved):
Looks better then what I was thinking. I removed the changes in my PR for updating the cron.

PR #139, opened 2025-01-08
Fix quests time, clickable gifts and confetti

valentinludu:
Changes:
- FE and BE both are considering UTC time.
- Fixed confetti
- Fixed 2 gift boxes available in the same time.

### DIFF
--- a/packages/scoutgame/src/claims/claimDailyRewardAction.ts
+++ b/packages/scoutgame/src/claims/claimDailyRewardAction.ts
@@ -12,8 +12,7 @@ export const claimDailyRewardAction = authActionClient
     const data = await claimDailyReward({
       userId: ctx.session.scoutId,
       isBonus: parsedInput.isBonus,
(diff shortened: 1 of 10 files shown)


packages/scoutgame/src/claims/claimDailyRewardAction.ts
```diff
@@ -12,8 +12,7 @@ export const claimDailyRewardAction = authActionClient
     const data = await claimDailyReward({
```
mattcasey, line 12: we should add validation now, since we have a deterministic timezone. if dayOfWeek is not today, it should throw an error

mattcasey (approved):
left one suggestion

packages/scoutgame/src/claims/claimDailyRewardAction.ts
valentinludu: Updated!

PR #142, opened 2025-01-08
Verify email

mattcasey:
- during onboarding, we add a description to check email and send the verification after you submit
- you can also see status and re-request verification from Accounts page
- adds a few extra fields to mixpanel
- instead of scoring referral points after onboarding, you have to verify email first so we check referrals whenever u verify an email
- check that no one else with similar emails has been referred yet
- adds a new 'sonner' library to manage toasts: https://sonner.emilkowal.ski/. This is very lightweight and easy-to-use, doesn't require managing state like MUI's Snackbar

image


.github/workflows/deploy_onchain_scoutgame.yml
```diff
@@ -123,7 +123,7 @@ jobs:
         run: npm run onchain-scoutgame:test
 
       - name: Run package tests
-        run: npx dotenv -e .env.test.local -- npm run test -w packages/scoutgame
```
mattcasey, line 126: now we'll test all package  tests, not just the scoutgame one

apps/cron/src/tasks/syncExternalUserProfiles/syncExternalUserProfilesTask.ts
```diff
@@ -31,6 +31,23 @@ async function getUsers({ offset = 0 }: { offset?: number } = {}): Promise<
         where: {
           type: 'referral'
         }
+      },
+      emailVerifications: {
```
mattcasey, line 35: update mixpanel records

packages/waitlist/package.json
```diff
@@ -4,7 +4,7 @@
   "type": "module",
   "devDependencies": {},
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": ""
```
mattcasey, line 7: we need a "test" command but don't want to show any output when the package tests get run

packages/scoutgame-ui/src/components/verify-email/VerifyEmailPage.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { useRouter } from 'next/navigation';
+import { useEffect } from 'react';
+import { toast } from 'sonner';
+
+const toastOptions = {
+  duration: 40000 // make it extra long so users can see it
+};
+
+export function VerifyEmailPage({ result }: { result: 'verified' | 'failed' | 'already_verified' }) {
+  const router = useRouter();
```
Devorein, line 12: We might need to use a ref to prevent it from being called twice

Devorein:
A small UX issue. I got two email verified toast instead of one.
image

PR #144, opened 2025-01-09
Improved resettable quests detection

Devorein:


apps/cron/src/tasks/processBuilderActivity/recordMergedPullRequest.ts
```diff
@@ -278,7 +278,9 @@ export async function recordMergedPullRequest({
               questTypes.push('contribute-celo-repo');
             }
 
-            await completeQuests(githubUser.builderId, questTypes);
+            if (questTypes.length) {
```
mattcasey, line 281: would be better if completeQuests just silently continue so you dont have to add this logic here. i think i saw it someplace else also

packages/scoutgame/src/points/builderEvents/sendPointsForSocialQuest.ts
```diff
@@ -1,24 +1,23 @@
...
 }: {
+  season: string;
   type: QuestType;
   builderId: string;
-  points: number;
   week: ISOWeek;
   tx?: Prisma.TransactionClient;
 }) {
-  const season = getCurrentSeasonStart(week);
+  const points = questsRecord[type].points;
```
mattcasey, line 20: why is this better than looking at the week? if week and season dont match then there is a problem

mattcasey (approved):

PR #146, opened 2025-01-10
Estimated payouts

motechFR:


packages/scoutgame/src/builderNfts/getAllSeasonNftsWithOwners.ts
```diff
@@ -0,0 +1,42 @@
...
+export type SeasonNfts = {
+  default: BuilderNftWithOwners[];
+  starter_pack: BuilderNftWithOwners[];
+};
+
+/**
+ * Get all the NFTs for a specific season along with their owners. This is based on the scoutNFT table which is synced with onchain activity ie. mints, transfers and burns
+ */
+export async function getAllSeasonNftsWithOwners({ season }: { season: ISOWeek }): Promise<SeasonNfts> {
+  const seasonBuilderNfts = await prisma.builderNft.findMany({
```
mattcasey, line 18: these queries could all be parallelized

packages/scoutgame/src/builderNfts/recordNftMint.ts
```diff
@@ -288,6 +289,13 @@ export async function recordNftMint(
     log.error('Error recording referral bonus', { error, builderId: builderNft.builderId, scoutId });
   }
 
+  refreshEstimatedPayouts({
```
mattcasey, line 292: i think this should block, since this is called from the webapp when a user buys an NFT. wouldn't we want the est price to be refreshed when we update the NFT view?

packages/scoutgame/src/builderNfts/recordNftMint.ts
```diff
@@ -288,6 +289,13 @@ export async function recordNftMint(
...
+    week: getCurrentWeek(),
+    builderIdToRefresh: builderNft.builderId
+  }).catch((error) => {
+    log.error('Error refreshing estimated payouts', { error, builderId: builderNft.builderId });
```
mattcasey, line 296: use userId, not builderId for logs

apps/cron/src/tasks/processBuilderActivity/index.ts
```diff
@@ -82,4 +83,8 @@ export async function processAllBuilderActivity(
   }
 
   await updateBuildersRank({ week: getCurrentWeek() });
+
+  await refreshEstimatedPayouts({ week: getCurrentWeek() }).catch((error) => {
+    log.error('Error refreshing estimated payouts', { error });
```
mattcasey, line 88: this error should bubble up so that the request fails and we catch it sooner

packages/scoutgame/src/builderNfts/refreshEstimatedPayouts.ts
```diff
@@ -0,0 +1,93 @@
...
+export async function refreshEstimatedPayouts({
+  week,
+  builderIdToRefresh
+}: {
+  week: ISOWeek;
+  builderIdToRefresh?: string;
+}): Promise<SeasonNfts> {
+  const season = getCurrentSeasonStart(week);
+
+  const { normalisedBuilders } = await getPointsCountForWeekWithNormalisation({
```
mattcasey, line 25: could be parallelized with the nfts query

packages/scoutgame/src/builderNfts/refreshEstimatedPayouts.ts
```diff
@@ -0,0 +1,93 @@
...
+          },
+          data: {
+            estimatedPayout: expectedPayoutForNextStarterPackPurchase
+          }
+        });
+      }
+    }
+  }
+
+  return getAllSeasonNftsWithOwners({ season });
```
mattcasey, line 92: this is probably a heavy method and there's no need for an update method to return data. the caller can just call this if they want

packages/testing/src/database.ts
```diff
@@ -26,18 +26,27 @@ export async function mockBuilder({
...
+  wallets = [],
+  weeklyStats = []
 }: Partial<
   Scout & {
     githubUserId?: number;
     githubUserLogin?: string;
     createNft?: boolean;
     nftSeason?: string;
     wallets?: { address: string }[];
+  } & {
```
mattcasey, line 38: why is there an "} & {" breaking these two inline interfaces up?

mattcasey: also just noting i came. up with a slightly different interface to pass in stats when using mockScout. not sure either is better or that they necessarily have to be the same

packages/scoutgame/src/builderNfts/refreshEstimatedPayouts.ts
```diff
@@ -0,0 +1,93 @@
...
+      const defaultNftBalance = defaultNft.nftOwners.reduce((acc, nft) => acc + nft.balance, 0);
+      const starterPackNftBalance = starterPackNft?.nftOwners.reduce((acc, nft) => acc + nft.balance, 0) ?? 0;
+
+      const builderNftBalanceWeighted = defaultNftBalance * nftTypeMultipliers.default;
+      const starterPackNftBalanceWeighted = starterPackNftBalance * nftTypeMultipliers.starter_pack;
+
+      const totalNftBalance = builderNftBalanceWeighted + starterPackNftBalanceWeighted;
+
+      const expectedPayoutForNextNftPurchase = Math.floor(
+        normalisedPoints / (totalNftBalance + nftTypeMultipliers.default)
```
mattcasey, line 61: maybe just a comment here to remind the reader why we have to look up nftTypeMultipliers, like "Use value from nftTypeMultipliers.default to increase the nft balance by one"?

The variable name multiplier doesnt mean anything to me and it's weird we are adding it

packages/scoutgame/src/builderNfts/refreshEstimatedPayouts.ts
motechFR: Noted & fixed

packages/scoutgame/src/builderNfts/getAllSeasonNftsWithOwners.ts
motechFR: Nice, fixed

packages/scoutgame/src/builderNfts/recordNftMint.ts
motechFR: 200ms on a prod connection, going to await it

packages/scoutgame/src/builderNfts/recordNftMint.ts
motechFR: Agreed & fixed. BuilderId as standard, and userId for the scout actually minting

packages/scoutgame/src/builderNfts/refreshEstimatedPayouts.ts
motechFR: Fixed

PR #150, opened 2025-01-11
Fix admin page

motechFR:


packages/safetransactions/tsconfig.json
mattcasey, line 1: a good case for a new package thanks :)

mattcasey (approved):

PR #152, opened 2025-01-13
use new collection to read NFT stats

mattcasey:
This PR uses the mutable ScoutNFT collection to determine what NFTs you own or have sold.

Some use cases where i left NFT Purchase Events:
- New Scouts table: to determine scouts that made their first NFT purchase
- Rewards on Claim page: to determine the source of points and how many NFTs you owned at the time
- Builder Activity, to show that an NFT was purchased as an event


packages/testing/src/generators.ts
```diff
@@ -1,12 +1,23 @@
...
+import type { Address } from 'viem';
+import { keccak256 } from 'viem/utils';
 
 export const randomLargeInt = () => Math.floor(Math.random() * 1000000000) + 1000000000;
 
 export function randomIntFromInterval(min: number, max: number) {
   return Math.floor(Math.random() * (max - min + 1) + min);
 }
 
+export function randomWalletAddress(): Address {
```
motechFR, line 14: Nice, keep needing random data

motechFR (approved):
Looks good, I noticed we don't have changes in cron here, but have an upcoming story to fix the refreshes of onchain data

Will be done inside

apps/cron/src/tasks/resolveMissingPurchases/findAndIndexMissingPurchases.ts

packages/testing/src/generators.ts
mattcasey: this was orginally a method you added i think, i just moved it into generators.ts since this is where i've been putting random utils for testing :)

PR #159, opened 2025-01-14
Cleanup tests

motechFR:


packages/testing/src/__tests__/deterministicBuildersData.spec.ts
```diff
@@ -0,0 +1,180 @@
+import { jest } from '@jest/globals';
+import { uuidFromNumber } from '@packages/utils/uuid';
```
mattcasey, line 2: you need to add @packages/utils as a dependency to @packages/testing

mattcasey (changes requested):
add utils to dependencies for the testing lib

mattcasey (approved):

PR #163, opened 2025-01-15
Scout AI agent for builders

motechFR:


packages/llm/src/constants.ts
```diff
@@ -0,0 +1,11 @@
+export const CHAT_GPT_BASE_URL = 'https://api.openai.com/v1/chat/completions';
+
+/**
+ * @see https://platform.openai.com/docs/models
+ */
+export const LLMModelsList = {
```
mattcasey, line 6: this mapping seems unecessary

packages/llm/src/tokenize.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Useful for checking context length
+ */
+export function tokenize({ text }: { text: string }): number {
+  const enc = encoding_for_model('gpt-4o');
+
+  const tokens = enc.encode(text);
+
+  enc.free();
```
mattcasey, line 12: what does this do?

packages/telegram/src/getTelegramChatHistory.ts
```diff
@@ -0,0 +1,46 @@
...
+  let skip = 0;
+
+  if (limit) {
+    const totalMessages = await prisma.agentTelegramMessage.count({
+      where: {
+        conversationTelegramId: Number(telegramChatId)
+      }
+    });
+    // If the total number of messages is less than the limit, set skip to 0
+    skip = Math.max(0, totalMessages - Number(limit));
```
mattcasey, line 29: its not clear why we would ever skip messages. The comment is confusing too since skip is 0 by default.

packages/telegram/src/getTelegramChatHistory.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const history = await prisma.agentTelegramMessage.findMany({
+    where: {
+      conversationTelegramId: Number(telegramChatId),
+      createdAt: {
+        gte: maxAgeInMinutes ? new Date(Date.now() - maxAgeInMinutes * 60 * 1000) : undefined
+      }
+    },
+    orderBy: {
+      createdAt: 'asc'
```
mattcasey, line 40: the function says "starting with most recent" then we should be sorting by descending

packages/telegram/src/getTelegramChatHistory.ts
mattcasey: Did AI write this? I think that this functino does give you the most recent comments , but it is out of order since it is giving oldest first. You should just order by date descending and pass the limit. No need to skip anything

packages/telegram/src/sendTelegramChatMessage.ts
```diff
@@ -0,0 +1,82 @@
...
+          : { is_disabled: true }
+      },
+      {
+        headers: {
+          'Content-Type': 'application/json'
+        }
+      }
+    );
+
+    return response as SendTelegramMessageResult;
```
mattcasey, line 63: you could pass the type in as a generic: `POST<SendTelegramMessageResult>`

packages/telegram/src/setTelegramBotWebhook.ts
```diff
@@ -0,0 +1,20 @@
...
+import type { BotToken } from './auth';
+
+/**
+ * Set the webhook for a telegram bot
+ */
+export async function setTelegramBotWebhook({ token, endpoint }: BotToken & { endpoint: string }) {
+  const url = `${getTelegramBaseUrl({ token })}/setWebhook?url=${encodeURIComponent(endpoint)}`;
+
+  await POST(url, undefined, {
+    query: {
```
mattcasey, line 13: you're putting the query in `url` but also pasing it in here, i'm assuming that we end up either adding two query strings or the POST is just de-duping it for you

packages/telegram/src/types.ts
mattcasey, line 1: interfaces.ts

scripts/generatePackage.sh
```diff
@@ -0,0 +1,78 @@
...
+EOL
+
+# Run eslint initialization
+cd "packages/$package_name"
+npx eslint --fix .
+cd ../..
+
+echo -e "Refreshing dependencies..."
+
+npm install
```
mattcasey, line 75: we should never just run npm install

packages/llm/src/constants.ts
motechFR: The goal here is threefold
- not tie ourselves to openAI conventions by leading model names throughout DB
- ability to pin a model to specific version, especially once we have something working
- prevent accidental usage of a more expensive model such as o1

scripts/generatePackage.sh
```diff
@@ -0,0 +1,78 @@
...
+  "exclude": ["jest.config.ts", "jest.teardown.ts"]
+}
+
+EOL
+
+# Create package.json
+cat > "packages/$package_name/package.json" << EOL
+{
+  "name": "@packages/$package_name",
+  "version": "0.0.1",
```
mattcasey, line 43: should be "0.0.0"

packages/llm/src/tokenize.ts
motechFR: This was in app.charmverse.io. I moved it over so we can use it. It allows us to tokenize text and get the same token counts as in ChatGPT

This allows us to simulate context window size to ensure we don't exceed it, and also predict pricing for processing a certain size of corpus of data (tokens * input tokens price)

scripts/generatePackage.sh
```diff
@@ -0,0 +1,78 @@
...
+  "private": true,
+  "main": "./index.ts",
+  "scripts": {
+    "test": "npx jest"
```
mattcasey, line 47: should be an empty string unless there actually are tests

packages/llm/src/types.ts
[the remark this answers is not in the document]
motechFR: Fair, will use this convention going forward

packages/telegram/src/getTelegramChatHistory.ts
motechFR: There is a mistake in the description.

It is purposefully in ascending order since we want the consumer of the method to be able to pass the messages as is to an AI, without additional sorting.

motechFR: The reason we want to skip old messages is for context window trimming. We don't want to always pass the full history (especially with cheaper models)

This method should remain unopiniated about the max context size, but provide controls to the call to define how many recent messages they want

apps/scout-agents/src/__tests__/healthCheck.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+
+afterAll((done) => {
+  server.close(done);
+});
+
+describe('test', () => {
+  test('should return 200', async () => {
+    // const response = await request(server).get(`/api/health`);
+    // expect(response.status).toBe(200);
+    // TODO: Fix local test runner
```
mattcasey, line 21: remember to clean this up or just delete the test ,,

scripts/generatePackage.sh
motechFR: This is what I usually do manually after creating a package. However, I'll make it npm install <package-name>

apps/scout-agents/src/__tests__/healthCheck.spec.ts
motechFR: I left this as a reminder as I'm having issues with jest. Keeps giving me an error with the local runner

I definitely want to have at least 1 working test

apps/scout-agents/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "@koa/router": "^13.1.0",
+    "@octokit/plugin-paginate-rest": "^11.3.5",
+    "@octokit/plugin-throttling": "^9.3.1",
+    "@packages/dates": "^0.0.0",
+    "@packages/farcaster": "^0.0.0",
+    "@packages/github": "^0.0.0",
+    "@packages/llm": "^0.0.1",
+    "@packages/mailer": "^0.0.0",
+    "@packages/mixpanel": "^0.0.0",
+    "@packages/moxie": "^0.0.0",
```
mattcasey, line 26: please remove packages we are not using here

apps/scout-agents/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "scout-agents",
+  "version": "1.0.0",
+  "description": "",
+  "type": "module",
+  "scripts": {
+    "build": "npx tsx esbuild.config.ts",
+    "start": "npx tsx watch src/main.ts",
+    "start:prod": "node --import dd-trace/register.js ./dist/main.js",
```
mattcasey, line 9: we dont need dd-trace for this app

apps/scout-agents/src/datadog.ts
```diff
@@ -0,0 +1,12 @@
+import tracer from 'dd-trace';
+
+if (process.env.NODE_ENV === 'production') {
+  tracer.init({
```
mattcasey, line 4: remove datadog APM. its expensive to run and not necessary to monitor this app

apps/scout-agents/src/main.ts
```diff
@@ -0,0 +1,39 @@
...
+        ]
+      }
+    });
+    log.info('Telegram bot webhook set');
+  } catch (error) {
+    log.error('Error setting telegram bot webhook', { error });
+  }
+}
+
+setupTelegramWebhook();
```
mattcasey, line 39: kinda hacky that this runs on every app start

apps/scout-agents/.ebextensions/00_env_vars.config
```diff
@@ -0,0 +1,13 @@
...
+# This file should only have
+# - env variables referenced in docker-compose file
+# - variables needed for the .platform/hooks/predeploy/01_pull_secrets.sh script to run
+
+option_settings:
+    aws:elasticbeanstalk:application:environment:
+        COMPOSE_PROJECT_NAME: "prd"
+        COMPOSE_PROFILES: "prd-scout-agents"
+        SERVICE_ENV: "prd"
+        SERVICE_NAME: "scout-agents"
```
mattcasey, line 11: remember we dont use hyphens in the service name. it should be scoutgameagents for consistency sake

apps/scout-agents/src/main.ts
motechFR: The core telegram bot identifier is the API token.

Setting the webhook is idempotent, so if you send the the same value as current it's totally fine.

I've put this here for simplicity so we can just run the app and have it setup telegram without extra config

PR #170, opened 2025-01-15
Fix calculate builder levels

motechFR:


packages/scoutgame/src/points/__tests__/calculateBuilderLevel.spec.ts
```diff
@@ -55,34 +59,39 @@ function validateCalculations({
...
+    averageGemsPerWeek: 666,
     centile: 23,
     level: 3,
-    firstActiveWeek: '2025-W07'
+    firstActiveWeek: '2025-W02'
   });
 }
 
-describe('calculateBuilderLevels', () => {
+describe.skip('calculateBuilderLevels', () => {
```
mattcasey, line 87: just fyi - is it intentional to skip this? r u planning on fixing later?

packages/scoutgame/src/points/__tests__/refreshBuilderLevels.spec.ts
```diff
@@ -1,39 +1,95 @@
...
+      amount: 200
+    });
+
+    await writeSeededBuildersGemPayoutsToDatabase({ builders, season });
+
+    const builder0 = builders[0];
+    const builder27 = builders[27];
+    const builder170 = builders[170];
+
+    prettyPrint({ builder0, builder27, builder170 });
```
mattcasey, line 59: remove log

packages/scoutgame/src/points/calculateBuilderLevel.ts
```diff
@@ -24,12 +24,23 @@ export const decileTable = [
...
-  season?: ISOWeek;
-}): Promise<BuilderAggregateScore[]> {
-  const allSeasonWeeks = getAllISOWeeksFromSeasonStart({ season });
+export async function calculateBuilderLevels(
+  {
+    season = getCurrentSeasonStart()
+  }: {
+    season?: ISOWeek;
+  } = {
+    season: getCurrentSeasonStart()
```
mattcasey, line 33: i dont think you need to set the default here, you can just do `= {}`

packages/scoutgame/src/points/calculateBuilderLevel.ts
```diff
@@ -39,6 +50,10 @@ export async function calculateBuilderLevels({
       },
       builderEvent: {
         season,
+        week: {
+          in: allSeasonWeeks
```
mattcasey, line 54: instead of calculating all weeks, just do "not this week" or "not greater or equal to this week".

packages/scoutgame/src/points/refreshBuilderLevels.ts
```diff
@@ -8,6 +8,8 @@ import { calculateBuilderLevels } from './calculateBuilderLevel';
 export async function refreshBuilderLevels({ season = getCurrentSeasonStart() }: { season?: ISOWeek }) {
   const levels = await calculateBuilderLevels({ season });
 
+  log.info(`Refreshing builder levels for season ${season} with ${levels.length} builders`);
```
mattcasey, line 11: this is a side effect so log.debug would be better

packages/scoutgame/src/points/refreshBuilderLevels.ts
```diff
@@ -26,5 +28,19 @@ export async function refreshBuilderLevels({ season = getCurrentSeasonStart() }:
...
+      userId: {
+        notIn: levels.map((level) => level.builderId)
+      }
+    },
+    data: {
+      level: 0
+    }
+  });
+
+  log.info(`Refreshed builder levels for season ${season} with ${levelZeroBuilders.count} unranked builders`);
```
mattcasey, line 45: we dont need three info logs every time this happens. you should at least remove the log in the middle so our logs arent very noisy

packages/scoutgame/src/points/calculateBuilderLevel.ts
```diff
@@ -71,7 +92,7 @@ export async function calculateBuilderLevels({
         acc[builderId] = {
           builderId,
           totalPoints: 0,
-          firstActiveWeek: getWeekFromDate(receipt.createdAt),
+          firstActiveWeek: receipt.week,
```
mattcasey, line 95: nice cleanup

packages/scoutgame/src/points/__tests__/refreshBuilderLevels.spec.ts
```diff
@@ -1,39 +1,95 @@
...
 
 import { refreshBuilderLevels } from '../refreshBuilderLevels';
 
+jest.useFakeTimers();
+
 describe.skip('refreshBuilderLevels', () => {
-  it('should refresh builder levels in the database', async () => {
-    const { builders, weeks } = seedBuilders({ season: getPreviousSeason(getCurrentSeasonStart()) });
+  beforeEach(() => {
+    jest.setSystemTime(new Date('2025-01-25'));
```
mattcasey, line 17: I don't want us to start using this. pattern. please change the methods to pass in the dates properly. for example u can pass in "now"

packages/scoutgame/src/points/__tests__/calculateBuilderLevel.spec.ts
motechFR: Fixed what caused tests to interfere with each other, unskipped

packages/scoutgame/src/points/__tests__/refreshBuilderLevels.spec.ts
motechFR: done

packages/scoutgame/src/points/calculateBuilderLevel.ts
motechFR: Nice, thanks

packages/scoutgame/src/points/refreshBuilderLevels.ts
motechFR: I grouped these into a single log.info

packages/scoutgame/src/points/refreshBuilderLevels.ts
```diff
@@ -26,5 +26,20 @@ export async function refreshBuilderLevels({ season = getCurrentSeasonStart() }:
       }
     });
   }
-  log.info(`Refreshed builder levels for season ${season} with ${levels.length} builders`);
+
+  const levelZeroBuilders = await prisma.userSeasonStats.updateMany({
```
mattcasey, line 30: in what situation do we need to set builders back to level 0?

mattcasey (approved):
Approved but i would like to remove two things that do not seem necessary:
1) theres no need to filter out gem receipts for current date
2) no need to set builders back to 0

Both of these seem like 'just-in-case' code which ultimately makes our system harder to reason about

packages/scoutgame/src/points/refreshBuilderLevels.ts
motechFR: Thanks, this is a one off situation and should actually be in the script

PR #175, opened 2025-01-16
Auto approve/reject builders

valentinludu:


Devorein (approved):

apps/cron/src/tasks/processBuilderActivity/reviewBuildersStatus.ts
```diff
@@ -0,0 +1,55 @@
...
+      // Get the activity of the builder in the last 28 days
+      const { commits, pullRequests } = await getBuilderActivity({
+        login: githubUser.login,
+        githubUserId: githubUser.id,
+        after: last28Days
+      });
+
+      // If the builder has activity in the last 28 days approve it
+      if (commits.length > 0 || pullRequests.length > 0) {
+        await prisma.scout.update({
```
Devorein, line 34: Rather than directly updating the builderStatus we should call this function `packages/scoutgame/src/builders/approveBuilder.ts` it will take care of creating the builder nft too.

Devorein (changes requested):

apps/cron/src/tasks/processBuilderActivity/reviewBuildersStatus.ts
valentinludu: Yep. Missed it.

I updated it now.

Devorein (approved):

mattcasey (approved):

PR #183, opened 2025-01-20
Recording nft transfer

motechFR:


apps/cron/src/tasks/resolveMissingPurchases/findAndIndexMissingPurchases.ts
```diff
@@ -94,12 +99,17 @@ export async function findAndIndexMissingPurchases({
...
       scoutgameMintsLogger.error('Missing tx', missingTx.transactionHash, 'tokenId', key);
 
-      if (missingTx.args.to === '0x0000000000000000000000000000000000000000') {
-        scoutgameMintsLogger.error('Skipping burn tx', missingTx.transactionHash, 'tokenId', key);
-        // eslint-disable-next-line no-continue
-        continue;
-      } else if (missingTx.args.from !== '0x0000000000000000000000000000000000000000') {
-        scoutgameMintsLogger.error('Skipping transfer tx', missingTx.transactionHash, 'tokenId', key);
+      // Null to means this is a burn, which impacts the total supply. Not null from means this is a transfer from an existing wallet
+      if (missingTx.args.to === NULL_EVM_ADDRESS || missingTx.args.from !== NULL_EVM_ADDRESS) {
```
mattcasey, line 103: so what is the case we are not recording an NFT transfer? it's when from is NULL? is that when someone first purchases a token?

packages/scoutgame/src/builderNfts/accounting/getTransferSingleEvents.ts
```diff
@@ -48,11 +49,15 @@ export function getTransferSingleEvents({
     );
 }
 
-export function getStarterPackTransferSingleEvents({ fromBlock, toBlock }: BlockRange): Promise<TransferSingleEvent[]> {
+export function getStarterPackTransferSingleEvents({
+  fromBlock,
+  toBlock,
+  season
+}: BlockRange & { season?: ISOWeek }): Promise<TransferSingleEvent[]> {
```
mattcasey, line 56: suggestion: make season required on low-level methods like these. The caller should always know which contract it's asking about, including which season

packages/scoutgame/src/builderNfts/constants.ts
```diff
@@ -57,8 +57,16 @@ export function getBuilderNftStarterPackContractAddress(season: ISOWeek = getCur
...
 export const MAX_STARTER_PACK_PURCHASES = 3;
 
-export function getBuilderNftContractAddressForNftType(nftType: BuilderNftType): Address {
-  return nftType === 'starter_pack' ? getBuilderNftStarterPackContractAddress() : getBuilderNftContractAddress();
+export function getBuilderNftContractAddressForNftType({
+  nftType,
+  season
+}: {
+  nftType: BuilderNftType;
+  season?: ISOWeek;
```
mattcasey, line 65: suggestion: make this required

packages/scoutgame/src/builderNfts/recordNftTransfer.ts
```diff
@@ -0,0 +1,78 @@
...
+  to,
+  tokenId,
+  txHash
+}: RecordNftTransferParams): Promise<void> {
+  const matchingNft = await prisma.builderNft.findFirstOrThrow({
+    where: {
+      tokenId,
+      contractAddress: {
+        equals: contractAddress,
+        mode: 'insensitive'
```
mattcasey, line 28: do we not lowercase contract addresses?

packages/scoutgame/src/builderNfts/recordNftTransfer.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+  });
+
+  await Promise.all([
+    refreshScoutNftBalance({
```
mattcasey, line 34: shouldn't we refresh balances after registering the new NFT event?

packages/scoutgame/src/builderNfts/recordNftTransfer.ts
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  if (existingNftPurchaseEvent) {
+    return;
+  }
+
+  await prisma.nFTPurchaseEvent.create({
+    data: {
+      // An NFT transferred in the secondary market does not generate points
+      pointsValue: 0,
```
mattcasey, line 70: do we have a date of the transaction? that should be applied to 'createdAt' on the purchase event

packages/scoutgame/src/builderNfts/recordNftTransfer.ts
```diff
@@ -0,0 +1,78 @@
...
+      tokenId,
+      contractAddress,
+      nftType: 'default'
+    })
+  ]);
+
+  const existingNftPurchaseEvent = await prisma.nFTPurchaseEvent.findFirst({
+    where: {
+      builderNftId: matchingNft.id,
+      txHash,
```
mattcasey, line 51: isnt txHash enough to look for an existing event? if there was an nft even w/the same txhash but different props, should that log an error?

packages/scoutgame/src/builderNfts/refreshScoutNftBalance.ts
```diff
@@ -0,0 +1,64 @@
...
+      tokenId,
+      contractAddress: {
+        equals: contractAddress,
+        mode: 'insensitive'
+      }
+    }
+  });
+
+  const balance = await (nftType === 'starter_pack'
+    ? getBuilderNftStarterPackReadonlyClient({
```
mattcasey, line 32: suggestion maybe for later: we could probably just one one simple client interface since we just need to call getBalance. you could probably use a standard client for ERC1155 even.

mattcasey (approved):
overall looks good, i like the cleanup to remove scoutId from nftpurchase events. i see you're just using null for mints still... this seems to be most intuitive for me.

PR #184, opened 2025-01-20
Merge 2 apps

valentinludu:


packages/scoutgame-ui/src/components/welcome/how-it-works/HowItWorksContent.tsx
```diff
@@ -1,14 +1,18 @@
...
 import { PointsIcon } from '@packages/scoutgame-ui/components/common/Icons';
 import { useMdScreen } from '@packages/scoutgame-ui/hooks/useMediaScreens';
 import Link from 'next/link';
 import React from 'react';
 
 export function HowItWorksContent({ onClickContinue }: { onClickContinue?: React.MouseEventHandler }) {
   const isMdScreen = useMdScreen();
+  const platform = getPlatform();
   const iconSize = isMdScreen ? 24 : 18;
+  const url = platform === 'telegram' ? '/quests' : '/builders-you-know';
```
mattcasey, line 14: We should remove this check. /builders-you-know already is responsible to do a redirect if you have no farcaster friends, and it also checks if you have farcaster

packages/nextjs/src/session/getSession.ts
```diff
@@ -1,3 +1,4 @@
+import { getPlatform } from '@packages/mixpanel/platform';
```
mattcasey, line 1: you need to add packages/mixpanel to the dependencies of this lib when you import it

packages/mixpanel/src/platform.ts
```diff
@@ -0,0 +1,18 @@
+import env from '@beam-australia/react-env';
+import { ReferralPlatform } from '@charmverse/core/prisma';
+
+const platform = env('SCOUTGAME_PLATFORM') || process.env.REACT_APP_SCOUTGAME_PLATFORM;
+
+function isPlatform(_platform: string = ''): _platform is ReferralPlatform {
```
mattcasey, line 6: we really ought to just put this in packages/utils/constants or something less opinionated than mixpanel, but this is a good step in that direction

apps/scoutgame/middleware.ts
```diff
@@ -10,18 +11,25 @@ export async function middleware(request: NextRequest) {
...
+    console.log('Redirecting to login', { path, platform, ...session });
     return NextResponse.redirect(new URL('/login', request.url));
   }
 
   // Redirect to home if logged in user clicks on login
-  if (isLoggedIn && (path === '/login' || path === '/')) {
+  if (isLoggedIn && (path === '/login' || path === '/' || path === '/home')) {
     // eslint-disable-next-line no-console
-    console.log('Redirecting to home page', session);
+    console.log('Redirecting to scout page', { path, platform, session });
```
mattcasey, line 32: thx for cleaning up the logs

apps/scoutgame/app/robots.ts
```diff
@@ -1,10 +1,15 @@
+import { getPlatform } from '@packages/mixpanel/platform';
 import type { MetadataRoute } from 'next';
 
 export default function robots(): MetadataRoute.Robots {
+  const platform = getPlatform();
+  const isTelegram = platform === 'telegram';
```
mattcasey, line 6: suggestion: call this variable "disallowRobots"

mattcasey (approved):
some small comments

packages/scoutgame-ui/src/components/welcome/how-it-works/HowItWorksContent.tsx
valentinludu: Done

packages/nextjs/src/session/getSession.ts
valentinludu: We have that in package.json

apps/scoutgame/app/robots.ts
valentinludu: yeah, better. Done

PR #213, opened 2025-01-23
Allow merging multiple pull request per repo per day

Devorein:


apps/cron/src/tasks/processBuilderActivity/__tests__/recordMergedPullRequest.spec.ts
```diff
@@ -308,51 +308,7 @@ describe('recordMergedPullRequest', () => {
     expect(scoutActivities).toBe(1);
   });
 
-  it('should only create one builder event per repo per day', async () => {
```
mattcasey, line 311: instead of deleting this test, because this code is so important, can you change the expectations that both events are created when they are for the same repo?

PR #218, opened 2025-01-24
Fixes to social quests for farcaster and telegram

mattcasey:
- added completeQuest when signing up via farcaster or telegram
- abuse fix: move the completeQuest logic so that it only applies if we are actually adding farcaster or telegram to the user
- connecting an account that's already connected shouldnt show an error to the user, so i removed the thrown error


packages/scoutgame/src/farcaster/connectFarcasterAccount.ts
```diff
@@ -26,4 +29,8 @@ export async function connectFarcasterAccount({ fid, userId }: { fid: number; us
     where: { id: userId },
     data: { farcasterId: fid, farcasterName: profile.username }
   });
+
+  await completeQuests(userId, ['link-farcaster-account']);
```
mattcasey, line 33: the bug was that completeQuests would be called even if we returning a 'connected user'

PR #219, opened 2025-01-24
hot fix for calculating PR window streak

mattcasey:


apps/cron/src/tasks/processBuilderActivity/recordMergedPullRequest.ts
```diff
@@ -56,7 +56,7 @@ export async function recordMergedPullRequest({
       createdBy: pullRequest.author.id,
       // streaks are based on merged date
       completedAt: {
-        gte: new Date(new Date(pullRequest.createdAt).getTime() - streakWindow)
+        gte: new Date(pullRequestDate.getTime() - streakWindow)
```
mattcasey, line 59: should be mergedAt, not createdAt

PR #244, opened 2025-01-29
Update scout wallet

valentinludu:
Add primary to scout wallets
Delete wallets


packages/scoutgame/src/wallets/updatePrimaryWallet.ts
```diff
@@ -0,0 +1,25 @@
+import { prisma } from '@charmverse/core/prisma-client';
+import { getAddress } from 'viem/utils';
+
+export async function updatePrimaryWallet(_address: string, userId: string) {
+  const address = getAddress(_address).toLowerCase();
+
+  const userWallet = await prisma.scoutWallet.findUnique({
```
mattcasey, line 7: better to use a count than findUnique, since findUnique retursn data we dont need, and can break if we do a db migration and the schema doesnt line up

packages/scoutgame-ui/src/components/accounts/components/WalletConnect.tsx
```diff
@@ -54,11 +59,32 @@ function WalletConnectButton({ user }: { user: UserWithAccountsDetails }) {
...
+      }
+    }
+  );
+
+  const { executeAsync: deleteWallet, isExecuting: isDeletingWallet } = useAction(deleteWalletAction, {
+    onSuccess: () => {
+      toast.success('Wallet deleted successfully');
+    },
+    onError: () => {
+      toast.error('Failed to delete wallet');
```
mattcasey, line 84: would to include the error message here

mattcasey (approved):
looks very thorough thanks

packages/scoutgame/src/wallets/updatePrimaryWallet.ts
valentinludu: Agree.

packages/scoutgame-ui/src/components/accounts/components/WalletConnect.tsx
valentinludu: Done.

PR #253, opened 2025-01-30
Better toast ux

valentinludu:


mattcasey (approved):

packages/scoutgame-ui/src/components/common/NFTPurchaseDialog/components/NFTPurchaseForm.tsx
```diff
@@ -172,9 +171,13 @@ export function NFTPurchaseFormContent({ builder }: NFTPurchaseProps) {
     onError({ error, input }) {
       scoutgameMintsLogger.error('Error purchasing with points', { input, error, userId: user?.id });
       setSubmitError(error.serverError?.message || 'Something went wrong');
+      toast.success('Error while purchasing with points');
```
mattcasey, line 174: i dont think we should show a toast here, there. is already an error inline in the form

packages/scoutgame-ui/src/providers/PurchaseProvider.tsx
```diff
@@ -145,6 +144,7 @@ export function PurchaseProvider({ children }: { children: ReactNode }) {
         {
           onSuccess: async (_data) => {
             setPurchaseSuccess(true);
+            toast.info('NFT purchase is sent and will be confirmed shortly');
```
mattcasey, line 147: we already show the user an error inline in the form when sendMintTransaction fails, so i dont think we need a message here

packages/scoutgame-ui/src/providers/PurchaseProvider.tsx
```diff
@@ -66,11 +65,11 @@ export function PurchaseProvider({ children }: { children: ReactNode }) {
     executeAsync: checkDecentTransaction
   } = useAction(checkDecentTransactionAction, {
     onSuccess({ input }) {
-      showMessage(`Transaction ${input.txHash || ''} was successful`, 'success');
+      toast.success(`Transaction ${input.txHash || ''} was successful`);
```
mattcasey, line 68: ideally, I thought we'd use sonner's loading toast to show the state here: https://sonner.emilkowal.ski/toast#promise. I think to do it properly, you would capture the promise returned from checkDecentTransaction. You could wrap the method inside of here so the components don't have to do it:
```
const checkDecentTransactionWrapped = (args: any) => {
  const decentTxPromise =  checkDecentTransaction(args);
  toast.promise(decentTxPromise, {
    loading: 'Loading...',
    success: (data) => {
      return `${data.name} toast has been added`;
    },
    error: 'Error',
  });
});
``

**Review by mattcasey — CHANGES_REQUESTED** [review]
Please remove toasts in the two places i noted that we already have a message in the form. i dont think we should start adding toasts everywhere just yet.. :)

And I had a suggestion to use a loading toast for the Decent transaction.

File: packages/scoutgame-ui/src/components/common/NFTPurchaseDialog/components/NFTPurchaseForm.tsx (hunk 1 above)
Comment (valentinludu) [reply packages/scoutgame-ui/src/components/common/NFTPurchaseDialog/components/NFTPurchaseForm.tsx:174]: Done

File: packages/scoutgame-ui/src/providers/PurchaseProvider.tsx (hunk 3 above)
Comment (valentinludu) [reply packages/scoutgame-ui/src/providers/PurchaseProvider.tsx]: Good catch. I didn't know we have that capability.

I thought of a simple implementation which I pushed:

```
const checkResultPromise = checkDecentTransaction({
    pendingTransactionId: res.data.id,
    txHash: res.data.txHash
});

toast.promise(checkResultPromise, {
  loading: 'Transaction is being settled...',
  success: () => `Transaction ${res?.data?.txHash || ''} was successful`,
  error: (data) => `Transaction failed: ${data?.serverError?.message || 'Something went wrong'}`
});

const checkResult = await checkResultPromise;

.... Rest is the same
```

mattcasey (approved):
sweet thanks for the updates

packages/scoutgame-ui/src/providers/PurchaseProvider.tsx
mattcasey: lgtm

PR #259, opened 2025-01-31
Added octant partner rewards page

Devorein:


valentinludu (approved):
LGTM!

apps/scoutgame/public/images/crypto/octant.svg
mattcasey, line 1: not a big deal but octant is not crypto, its a startup incubator

PR #265, opened 2025-02-03
Send email to user if tx fails

valentinludu:


packages/scoutgame/src/builderNfts/handleEmailFailedTransaction.ts
```diff
@@ -0,0 +1,49 @@
...
+      where: {
+        id: pendingTransactionId
+      }
+    });
+
+    if (!pendingTx) {
+      return;
+    }
+
+    const user = await getUser(userId);
```
Devorein, line 26: its not necessary as we already fetch the user inside `sendEmailTemplate`

Devorein (approved):

packages/scoutgame/src/builderNfts/handleEmailFailedTransaction.ts
valentinludu: I need it for the `name` variable

PR #271, opened 2025-02-05
Token airdrop claim

Devorein:


packages/scoutgame-ui/src/components/claim/components/PointsTable/ClaimedPointsTable.tsx
```diff
@@ -17,10 +19,17 @@ export async function ClaimedPointsTable() {
     isClaimed: true
   });
 
+  const partnerRewards = await getPartnerRewards({
```
mattcasey, line 22: you should get this in parallel with pointsReceipts so it runs faster

packages/scoutgame-ui/src/components/claim/components/PointsTable/PointsTable.tsx
```diff
@@ -1,23 +1,62 @@
...
+import type { PartnerReward } from '@packages/scoutgame/points/getPartnerRewards';
 import type { PointsReceiptReward } from '@packages/scoutgame/points/getPointsReceiptsRewards';
-import type { ReactNode } from 'react';
+import { useMemo, type ReactNode } from 'react';
 
 import { StyledTableBody, StyledTableHead } from '../common/StyledTable';
 
 import { PointsReceiptRewardRow } from './PointsReceiptRewardRow';
 
+const getTypeOrder = (type: string): number => {
```
mattcasey, line 12: I think a list would work better here? 
```
const getTypeOrder = (type: string): number => types.indexOf(type)
```

packages/scoutgame-ui/src/components/claim/components/PointsTable/UnclaimedPointsTable.tsx
```diff
@@ -19,10 +21,19 @@ export async function UnclaimedPointsTable() {
     })
   );
 
+  const [, partnerRewards = []] = await safeAwaitSSRData(
```
mattcasey, line 24: I believe it would be better to let PointsTable accept an undefined value for partnerRewards. The reason is that this default value is being re-created on every render, so it will always look like a new value to PointsTable and cause it to re-render when anything else on the page changes

packages/scoutgame/src/points/getClaimableTokensWithSources.ts
```diff
@@ -151,6 +152,16 @@ export async function getClaimableTokensWithSources(userId: string): Promise<Unc
     }))
     .filter((proof) => proof.amount > 0 && proof.proofs.length > 0 && !claimedWeeks.includes(proof.week));
 
+  const partnerRewardPayoutCount = await prisma.partnerRewardPayout.count({
```
mattcasey, line 155: i just realized this method is doing a lot of db lookups in series, would be worth refactoring at some point, or at least trying to add new queries in parallel.

packages/scoutgame/src/points/getPartnerRewards.ts
```diff
@@ -0,0 +1,121 @@
...
+}) {
+  const optimismNewScoutPartnerRewards: OptimismNewScoutPartnerReward[] = [];
+  const optimismTopReferrerRewards: OptimismTopReferrerReward[] = [];
+
+  const partnerRewardPayouts = await prisma.partnerRewardPayout.findMany({
+    where: {
+      userId,
+      claimedAt: isClaimed ? { not: null } : { equals: null },
+      payoutContract: {
+        partner: {
```
mattcasey, line 70: why do we care which partner it was?

apps/cron/src/tasks/processGemsPayout/createTopReferrerRewardsContract.ts
```diff
@@ -0,0 +1,62 @@
...
+  await prisma.partnerRewardPayoutContract.create({
+    data: {
+      chainId: optimism.id,
+      contractAddress,
+      season,
+      week,
+      cid,
+      tokenAddress: optimismTokenAddress,
+      tokenDecimals: optimismTokenDecimals,
+      partner: 'optimism_top_referrer',
```
mattcasey, line 45: "optimism_top_referrer" is not a partner, it is the name of a reward. Any time we use "partner," we should limit it to the values in our bonus config.

mattcasey, line 45: at first i left a comment saying we shouldnt make 'optimism_top_referrer' as a partner, but we do use 'partner reward' to describe it in the figma, so i think it makes sense

apps/cron/src/tasks/processGemsPayout/index.ts
```diff
@@ -60,6 +63,19 @@ export async function processGemsPayout(ctx: Context, { now = DateTime.utc() }:
...
+    const [newScoutRewards, topReferrerRewards] = await Promise.all([
+      createNewScoutRewardsContract({ week, season }),
+      createTopReferrerRewardsContract()
+    ]);
+    log.info('Rewards contract deployed', {
+      newScoutRewards,
+      topReferrerRewards
+    });
+  } catch (error) {
+    log.error('Error deploying new scout contract', { error });
```
mattcasey, line 76: i would log week and season here too since those are inputs to one of the methods

packages/scoutgame-ui/src/components/claim/components/PointsClaimContainer.tsx
```diff
@@ -31,10 +32,18 @@ export async function PointsClaimContainer() {
...
+  const { bonusPartners, points, builders, repos, partnerRewardPayoutCount } = data;
 
   const claimData = isOnchainApp ? (data as UnclaimedTokensSource).claimData : undefined;
 
+  const [unclaimedPartnerRewardsErr, unclaimedPartnerRewards] = await safeAwaitSSRData(
+    getUnclaimedPartnerRewards({ userId: scoutId })
+  );
+
+  if (unclaimedPartnerRewardsErr) {
+    return null;
```
mattcasey, line 44: i dont think we should be returning nothing here?

packages/scoutgame-ui/src/components/claim/components/PointsClaimScreen/PartnerRewardClaimButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+}) {
+  return (
+    <LoadingButton
+      variant='contained'
+      color='primary'
+      sx={{
+        width: 'fit-content'
+      }}
+      loading={isExecuting}
+      data-test='claim-points-button'
```
mattcasey, line 20: should not re-use this data-test attribute

packages/scoutgame-ui/src/components/claim/components/PointsClaimScreen/PartnerRewardClaimButton.tsx
mattcasey, line 1: Not a required change but this looks almost identical to PointsClaimButton.tsx with a different label. instead of adding another component it would be better to just reuse PointsClaimButton but use the children prop to customize the text instead of hardcoding it.

packages/scoutgame/src/points/getPartnerRewards.ts
```diff
@@ -0,0 +1,121 @@
...
+        select: {
+          partner: true
+        }
+      }
+    }
+  });
+
+  return partnerRewards.map((payout) => ({
+    id: payout.id,
+    amount: payout.amount,
```
mattcasey, line 48: remember we need to convert the amount to something that's user-friendly. it's stored in gwei right, so at least convert using the correct decimals.

packages/scoutgame/src/points/getPartnerRewards.ts
```diff
@@ -0,0 +1,121 @@
...
+export async function getPartnerRewards({
+  userId,
+  isClaimed,
+  season
+}: {
+  userId: string;
+  isClaimed: boolean;
+  season: string;
+}) {
+  const optimismNewScoutPartnerRewards: OptimismNewScoutPartnerReward[] = [];
```
mattcasey, line 62: suggestion: you just need one array. in this function's return response you combine them together and mix up the results by sorting by points anyway

mattcasey (changes requested):
left comments

packages/scoutgame/src/points/getClaimableTokensWithSources.ts
Devorein: for now I can just add the new partner rewards payout query in parallel

packages/scoutgame-ui/src/components/claim/components/PointsClaimContainer.tsx
Devorein: `safeAwaitSSRData` always logs the error inside its onError handler.

packages/scoutgame-ui/src/components/claim/components/PointsClaimScreen/PartnerRewardClaimButton.tsx
Devorein: yeah you are right, initially I wanted to move the action over here with extra logic but since these are passed as props we can re use the same component. I will just rename it to `RewardsClaimButton`

apps/cron/src/tasks/processGemsPayout/createSablierAirdropContract.ts
```diff
@@ -0,0 +1,151 @@
...
+  const createdCampaignAddress = receipt.logs[0].address as `0x${string}`;
+
+  const { request: approveCampaignRequest } = await publicClient.simulateContract({
+    address: tokenAddress,
+    abi: erc20Abi,
+    functionName: 'approve',
+    args: [createdCampaignAddress, aggregateAmount],
+    account
+  });
+
```
motechFR, line 141: As discussed, the merkle doesn't pull funds it's manually prefunded.

So this needs to be a transfer call


Please also confirm mechanics for refunding after claims expire

packages/blockchain/src/airdrop/claimSablierAirdrop.ts
```diff
@@ -0,0 +1,105 @@
...
+  contractAddress,
+  recipientAddress,
+  cid,
+  adminPrivateKey
+}: {
+  chainId: number;
+  contractAddress: Address;
+  recipientAddress: Address;
+  cid: string;
+  adminPrivateKey: `0x${string}`;
```
motechFR, line 19: We shouldn't need an admin key.

I'm assuming anyone can pay for gas to make a claim for anyone.

So the test is to use a wallet that has a bit of ETH but is not admin or mentioned in the airdrop

apps/cron/src/tasks/processGemsPayout/deployNewScoutRewardsContract.ts
```diff
@@ -0,0 +1,66 @@
...
+    campaignName: `Scoutgame New Scout S${season}W${getCurrentSeasonWeekNumber(season)} Rewards`,
+    chainId: optimismSepolia.id,
+    recipients: top10Scouts.map((scout, index) => ({
+      address: scout.address as `0x${string}`,
+      amount: newScoutsRewards[index]
+    })),
+    tokenAddress: optimismTokenAddress,
+    tokenDecimals: optimismTokenDecimals
+  });
+
```
mattcasey, line 30: make sure to log the contract address here in case the db save fails

apps/cron/src/tasks/processGemsPayout/deployTopReferrerRewardsContract.ts
```diff
@@ -0,0 +1,92 @@
...
+import { DateTime } from 'luxon';
+import { parseUnits } from 'viem';
+import { optimismSepolia } from 'viem/chains';
+
+import { createSablierAirdropContract } from './createSablierAirdropContract';
+import { optimismTokenDecimals, optimismTokenAddress } from './deployNewScoutRewardsContract';
+
+const REFERRAL_CHAMPION_REWARD_AMOUNT = parseUnits('25', optimismTokenDecimals).toString();
+
+export async function deployTopReferrerRewardsContract() {
```
mattcasey, line 14: if only one person is receiving reward, send something small to NULL address (0x000....)

apps/cron/src/tasks/processGemsPayout/deployTopReferrerRewardsContract.ts
```diff
@@ -0,0 +1,92 @@
...
+  const topConnectors: { address: string; date: DateTime }[] = [];
+  const week = getLastWeek();
+  const season = getCurrentSeasonStart(week);
+
+  for (let day = 1; day <= 7; day++) {
```
mattcasey, line 19: use result from `getDateFromISOWeek(week)` instead of `DateTime.utc()`

```
const date = weekStart.plus({ days: day });
```

apps/cron/src/tasks/processGemsPayout/deployTopReferrerRewardsContract.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (topConnectors.length === 0) {
+    log.info('No top connectors found for the week', {
+      week,
+      season
+    });
+    return;
+  }
+
+  if (topConnectors.length === 1) {
+    log.info('Only one top connector found for the week.', {
```
mattcasey, line 37: if only one person is receiving reward, send something small to NULL address (0x000....)

apps/cron/src/tasks/processGemsPayout/index.ts
```diff
@@ -60,6 +63,15 @@ export async function processGemsPayout(ctx: Context, { now = DateTime.utc() }:
     }
   }
 
+  try {
+    const [newScoutRewards, topReferrerRewards] = await Promise.all([
+      deployNewScoutRewardsContract({ week, season }),
+      deployTopReferrerRewardsContract()
```
mattcasey, line 69: add a .catch() onto each method here

packages/blockchain/src/airdrop/checkSablierAirdropEligibility.ts
```diff
@@ -0,0 +1,101 @@
+import { MerkleTree } from 'merkletreejs';
+import type { Address } from 'viem';
+import { keccak256, encodeAbiParameters, parseAbiParameters } from 'viem';
+
+import { getPublicClient } from '../getPublicClient';
+
+// @ts-ignore
+import sablierMerkleInstantAbi from './SablierMerkleInstant.json' assert { type: 'json' };
```
mattcasey, line 8: maybe try "resolveJsonModule": true in tsconfig.base.json

packages/blockchain/src/airdrop/clawbackAirdropTokens.ts
```diff
@@ -0,0 +1,101 @@
...
+    const { request } = await publicClient.simulateContract({
+      address: contractAddress,
+      abi: sablierAirdropAbi,
+      functionName: 'clawback',
+      args: [recipientAddress, balance],
+      account: walletClient.account
+    });
+
+    const hash = await walletClient.writeContract(request);
+    await publicClient.waitForTransactionReceipt({ hash });
```
mattcasey, line 81: add an info log w/the hash

packages/scoutgame-ui/src/components/claim/components/PointsClaimScreen/PartnerRewardsClaimButton/useClaimPartnerReward.ts
```diff
@@ -0,0 +1,177 @@
...
+    const { request } = await publicClient.simulateContract({
+      address: contractAddress,
+      abi: sablierAirdropAbi,
+      functionName: 'claim',
+      args: [BigInt(index), recipientAddress, BigInt(amount), proof],
+      account: recipientAddress
+    });
+
+    const hash = await walletClient.writeContract(request);
+    await publicClient.waitForTransactionReceipt({ hash });
```
mattcasey, line 71: could use walletClient

packages/scoutgame-ui/src/components/claim/components/PointsClaimScreen/PartnerRewardsClaimButton/useClaimPartnerReward.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    return { hash };
+  } catch (error) {
+    if (error instanceof Error) {
```
mattcasey, line 75: add a log error

mattcasey (approved):

packages/blockchain/src/airdrop/checkSablierAirdropEligibility.ts
Devorein: didn't work, I had to manually include the json file in the src field of tsconfig.json for blockchain package

apps/cron/src/tasks/processGemsPayout/deployTopReferrerRewardsContract.ts
Devorein: tested and clawback works too

PR #275, opened 2025-02-07
update math for starter pack pool

mattcasey:
We now have 3 pools to divide builder payout:
20% - builder
70% - regular nft holders
10% - starter pack holders


packages/scoutgame/src/points/divideTokensBetweenBuilderAndHolders.ts
```diff
@@ -108,14 +103,44 @@ export function divideTokensBetweenBuilderAndHolders({
   };
 }
 
-// returs the percentage of the total tokens that the scout should receive
+// Returns the percentage of the total weekly rewards that a scout should receive
+// Scout share percent = 0.7 * (owned default NFT / total default NFTs) + 0.1 * (owned starter pack NFT / total starter pack NFTs)
 export function calculateRewardForScout({
+  builderPool = 0.2,
+  starterPackPool = 0.1,
```
motechFR, line 110: This introduces the issue of having a coefficient where before we just had a base points pool.

I suggest changing the argument here to allocatedPoints: number 

Then doing the internal pool splits inside the method.

mattcasey: maybe you mean the builderPool? The reason i bring that in is because i am not always considering the starter pack pool, since we haven't minted it for everyone. So we use the remaining percent to determine the proper pool for default nfts. seemed better than just hard-coding the default to either be .7 or .8

PR #290, opened 2025-02-11
paginate result from moxie token

mattcasey:


packages/nextjs/src/session/getSession.ts
```diff
@@ -12,5 +13,13 @@ export async function getSession<T extends object = SessionData>(cookieOptions?:
 
   const session = await getIronSession<T>(cookies(), options);
 
+  // allow for a user override in development
```
mattcasey, line 16: added this for remote debugging

PR #310, opened 2025-02-17
Improved new scouts sorting logic to include earliest purchase date and nfts purchased count

Devorein:
A user in telegram mentioned why they were not part of the new scouts competition. The previous sorting logic only depended on the points earned and now since most scouts get 6-10 points (by scouting starter packs) only. it was random and dependent on the user who was queried from db first. Thus we needed to add a few more layers of sorting in the logic to break the tie. I went with nfts purchased before earliest purchased date since you spent more but we can talk about it in our dev chat.

### DIFF
--- a/packages/scoutgame/src/scouts/getNewScouts.ts
+++ b/packages/scoutgame/src/scouts/getNewScouts.ts
@@ -13,6 +13,7 @@ export type NewScout = {
   pointsPredicted: number;
   buildersScouted: number;
   nftsHeld: number;
+  latestPurchasedAt: Date | undefined;
 };
 
 // look at gems instead of points for current week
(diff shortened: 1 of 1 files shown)


mattcasey:
thank..., i think first purchase would be mor in line with other ranking

packages/scoutgame/src/scouts/getNewScouts.ts
```diff
@@ -13,6 +13,7 @@ export type NewScout = {
   pointsPredicted: number;
   buildersScouted: number;
   nftsHeld: number;
+  latestPurchasedAt: Date | undefined;
```
mattcasey, line 16: should be earliestPurchasedAt

PR #316, opened 2025-02-19
better cursor UX for scout table headers

mattcasey:
only turn the mouse into a pointer on columns that can be sorted


Devorein:
@mattcasey we use rank as the default builder sort inside `apps/scoutgame/app/(general)/developers/page.tsx`

Devorein (approved):

mattcasey:
oh thanks @Devorein .. interesting because getBuilders, which feeds scouttable, does not support rank for builders!

PR #318, opened 2025-02-19
remove some repos from octant rewards

mattcasey:


Devorein:
@mattcasey Have we removed the `bonusPartner` field in all these repos in the `GithubRepo` records?

mattcasey:
not yet, i was just going to do that in a script offline

PR #320, opened 2025-02-19
Developer mini profile modal

Devorein:


packages/scoutgame-ui/src/components/common/Card/BuilderCard/BuilderCardNftDisplay.tsx
```diff
@@ -44,13 +48,12 @@ export function BuilderCardNftDisplay({
...
   const height = nftDisplaySize[size].height;
-
+  const { openModal } = useDeveloperInfoModal();
   return (
     <Box overflow='hidden' width={width} height={height} sx={{ backgroundColor: 'black.dark', borderRadius: '4px' }}>
       <CardActionArea
         disabled={disableProfileUrl}
-        LinkComponent={Link}
-        href={`/u/${path}`}
+        onClick={() => openModal(path)}
```
mattcasey, line 56: I think it would be nice if we kept the href attributes so you could still right-click and open in a new tab

packages/scoutgame-ui/src/providers/DeveloperInfoModalProvider.tsx
```diff
@@ -0,0 +1,81 @@
...
+export function DeveloperInfoModalProvider({ children }: { children: ReactNode }) {
+  const [isOpen, setIsOpen] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const [developer, setDeveloper] = useState<DeveloperInfo | null>(null);
+  const router = useRouter();
+
+  const openModal = useCallback((path: string) => {
+    setIsLoading(true);
+    setIsOpen(true);
+    getDeveloperInfo(path)
```
mattcasey, line 30: Can we cache this with SWR, so that when you open the modal again, it is instant?

packages/scoutgame-ui/src/components/common/DeveloperInfoModal/DeveloperInfoModal.tsx
```diff
@@ -0,0 +1,397 @@
...
+            md: 600
+          },
+          bgcolor: 'black.main',
+          borderRadius: 2
+        }
+      }}
+    >
+      <Stack
+        onClick={() => {
+          router.push(`/u/${developer.path}`);
```
mattcasey, line 82: Confusing UX, but i think we should see what chris/andy think? i accidentally kept closing this/navigating to the page. i think at minimum we should make the cursor 'pointer', or instead have a "View Profile" button on the bottom of the card.

mattcasey (changes requested):
i think it is worth have a better loading state, and left some other smaller suggestions

mattcasey:
also there's something wrong with the carousel 
image

packages/scoutgame-ui/src/components/common/Card/BuilderCard/BuilderCardNftDisplay.tsx
Devorein: good idea. updated

packages/scoutgame-ui/src/providers/DeveloperInfoModalProvider.tsx
Devorein: yeah it will be instant if we do that

PR #327, opened 2025-02-20
remove Top Connector Referral quest and replace with new OP reward

mattcasey:
- confirmed with alex and chris we no longer will reward points/tokens, just OP instead
- remove the sidebar component for Top Connector
- replace the admin export with rewards for the past week
- updated the sablier contract to give referree and referrer 5 OP each
- grab the accurate number of tokens earned


Devorein (approved):
looks good we need to remove the +50 points inside `packages/scoutgame/src/quests/questRecords.ts` too

PR #333, opened 2025-02-21
Scoutgame farcaster frames v2

Devorein:


mattcasey (approved):

mattcasey (commented):
use findOrCreateFarcasterUser() inside of findOrCreateWalletUser if we find a farcaster profile, than ensure we add the latest wallet to the profile, and skip the call to findOrCreateUser

PR #348, opened 2025-02-27
add announcement banner for referral feature

mattcasey:


Devorein:
image
System wide Banner is not visible in the home page due to the starry bg. I'm not sure whether we even need to show it in the homepage.

Devorein (changes requested):
the announcement banner is shown for a split second even if its hidden in each reload. We should hide it by default and then once we verify its not hidden in localstorage only then we should show it

mattcasey:
> the announcement banner is shown for a split second even if its hidden in each reload. We should hide it by default and then once we verify its not hidden in localstorage only then we should show it

thanks for checking.. i am not sure if it is possible,will have to play around with it. cuz there's no "loading" state afaict

PR #353, opened 2025-02-28
log warning for points evaluated to 0

Devorein:


packages/scoutgame/src/points/getPointsCountForWeekWithNormalisation.ts
```diff
@@ -44,7 +45,16 @@ export async function getPointsCountForWeekWithNormalisation({
...
-    throw new Error('Points evaluated to 0');
+    log.warn('Points evaluated to 0', {
+      week
+    });
+    return {
+      totalPoints: 0,
+      normalisationFactor: 0,
+      normalisedBuilders: [],
+      weeklyAllocatedPoints,
+      topWeeklyBuilders: leaderboard
```
mattcasey, line 56: Should this be an empty array? when we call the method in processGemsPayout, we loop thru all topWeeklyBuilders to process points. this way, we dont loop at all and it just keeps going..

Devorein: correct it will have unintended sideeffects. we are returning processed data from this function so rather than returning empty values better to just throw errors and log it with a warning (as its rarely triggered)

PR #355, opened 2025-02-28
Fixed broken emojis that can't be rendered in share image

Devorein:


packages/scoutgame/src/builderNfts/artwork/generateShareImage.tsx
```diff
@@ -8,6 +7,8 @@ import type { BuilderStats } from '../../builders/getBuilderStats';
 
 import { BuilderShareImage } from './components/BuilderShareImage';
 
+const blacklistedEmojis = ['⌐◨', '-◨', '◨-'];
```
mattcasey, line 10: good idea

mattcasey (approved):

PR #362, opened 2025-03-03
Fixed client side admin partner page error

Devorein:


packages/users/src/findOrCreateUser.ts
```diff
@@ -66,12 +66,12 @@ export async function findOrCreateUser({
     : undefined;
 
   const scout = await prisma.scout.findFirst({
-    where: farcasterId
-      ? { farcasterId }
-      : telegramId
-        ? { telegramId }
-        : lowercaseAddresses
-          ? { wallets: { some: { address: { in: lowercaseAddresses } } } }
+    where: lowercaseAddresses
```
Devorein, line 69: reverted back to previous order since we are already handling farcaster login

PR #369, opened 2025-03-05
Notification preferences UI

Devorein:


packages/scoutgame-ui/src/components/accounts/components/NotificationSettings.tsx
```diff
@@ -0,0 +1,112 @@
...
+            variant='contained'
+            color='primary'
+            type='submit'
+            loading={isExecuting}
+            disabled={isExecuting || !isDirty}
+            sx={{ width: 'fit-content' }}
+          >
+            Save
+          </LoadingButton>
+          <FormErrors errors={errors} />
```
mattcasey, line 107: nice work getting the Formerrors and everything hooked up properly for this form.. i don't think i would have even been able to remember it all

mattcasey (approved):

PR #379, opened 2025-03-09
Show transferred nfts in developer page

Devorein:
[Builder](https://scoutgame.xyz/u/watchcoin-301725d8)
ccarella is missing since he used secondary marketplace to purchase this devs nft

### DIFF
--- a/packages/scoutgame/src/builders/getBuilderScouts.ts
+++ b/packages/scoutgame/src/builders/getBuilderScouts.ts
@@ -4,8 +4,6 @@ import type { BasicUserInfo } from '@packages/users/interfaces';
 import { BasicUserInfoSelect } from '@packages/users/queries';
 import { isTruthy } from '@packages/utils/types';
 
-import { validMintNftPurchaseEvent } from '../builderNfts/constants';
-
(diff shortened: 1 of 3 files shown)


mattcasey (approved):

packages/scoutgame/src/builders/getBuilderScouts.ts
```diff
@@ -38,12 +40,15 @@ export async function getBuilderScouts(builderId: string): Promise<BuilderScouts
     }
   });
 
+  // TODO: Figure out why there are some events that don't have a scoutWallet
+  const validPurchaseEvents = nftPurchaseEvents.filter((e) => e.scoutWallet);
```
mattcasey, line 44: i would check to see if they have a senderWallet address? These may be 'burned' nft purchases. If we send them to NULL address, there is no scout wallet

mattcasey (approved):

packages/scoutgame/src/builders/getBuilderScouts.ts
Devorein: added a check to make sure we don't fetch burned nfts

PR #388, opened 2025-03-11
Fix cache issue in building app for CI + fix build error

mattcasey:
We wanted to update the /farcaster/.well-known.json file, but for some reason in production it keeps getting the old file.


apps/scoutgame/app/(general)/info/repositories/page.tsx
```diff
@@ -6,6 +6,8 @@ export const metadata: Metadata = {
   title: 'Repositories'
 };
 
+export const dynamic = 'force-dynamic';
```
mattcasey, line 9: hoping to fix this build error:
image

PR #395, opened 2025-03-17
fix some errors in cron

mattcasey:
- use one global client for all github clients (since it manages throttling internally, should reduce rate limit errors)
- reduce dune analytics to once per day. even this is not enough to support many wallets:
 
we hit our monthly credit limit of 2500 with Dune (I didn't see this limitation til now)
With just 3 wallets, running once an hour we hit the limit in less than 3 days. So free tier gives us about 143 requests for free per month
The bad news is that the first paid tier is not much better, 4000 credits per month. For $349/mo we can get 25k credits: https://dune.com/pricing

### DIFF
--- a/apps/cron/cron.yml
+++ b/apps/cron/cron.yml
@@ -58,5 +58,5 @@ cron:
 
   - name: 'process-dune-analytics'
     url: '/process-dune-analytics'
-    # every hour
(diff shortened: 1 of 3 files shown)


mattcasey:
optimistically merging because i accidentally commited half to main :/

PR #402, opened 2025-03-19
simplify the logic to calculate builder levels

mattcasey:
was reading the code for levels and the logic was spread throughout the function, making it hard to read

also this adds a way to retrieve levels for previous weeks for offline analytics


packages/scoutgame/src/points/calculateBuilderLevel.ts
```diff
@@ -81,38 +94,30 @@ export async function calculateBuilderLevels({
...
           firstActiveWeek: gemsPayout.week,
+          activeWeeks,
           centile: 0,
           level: 0,
           averageGemsPerWeek: 0
         };
       }
+
       acc[builderId].totalPoints += gemsPayout.points;
+      acc[builderId].averageGemsPerWeek = Math.floor(acc[builderId].totalPoints / acc[builderId].activeWeeks);
```
mattcasey, line 114: instead of looping thru builders again, just calculate the avg again each time we find another week with points

PR #404, opened 2025-03-19
App notifications page

Devorein:
image


apps/scoutgame/app/(general)/notifications/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  const status = searchParams.status ?? 'all';
+
+  const [error, data] = await safeAwaitSSRData(
+    Promise.all([getAppNotifications({ userId, status }), getAppNotificationCount({ userId })])
+  );
+
+  if (error) {
+    return null;
```
mattcasey, line 25: i think i've asked before, and may have forgotten, but why return null if an error occurs?

packages/scoutgame-ui/src/components/notifications/NotificationCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import { toggleAppNotification } from '../../actions/toggleAppNotification';
+import { useMdScreen } from '../../hooks/useMediaScreens';
+
+export function NotificationCard({ notification }: { notification: ScoutAppNotification }) {
+  const { execute, isExecuting } = useAction(toggleAppNotification);
+  const notificationType = notification.notificationType as keyof typeof AppNotificationTypesRecord;
+  const title = AppNotificationTypesRecord[notificationType].title;
+
+  const description =
```
mattcasey, line 17: i counted only 3 types that don't use a function. might be good to just make them all functions so this code is simpler

packages/scoutgame/src/notifications/appNotificationConstants.ts
```diff
@@ -0,0 +1,89 @@
+type Variables = {
```
mattcasey, line 1: suggest renaming this to "templateVariables" since that's all that's really in here

mattcasey:
can we use the cat dog here?
image

packages/scoutgame-ui/src/components/notifications/NotificationCard.tsx
Devorein: yeah good point

PR #433, opened 2025-03-26
fix sorting for developers by rank

mattcasey:


packages/scoutgame/src/builders/getDevelopersForTable.ts
```diff
@@ -343,7 +343,7 @@ export async function getDevelopersForTable({
       avatar: builder.avatar as string,
       displayName: builder.displayName,
       // TODO: use the currentPriceInScoutToken when we move to $SCOUT
-      price: currentPrice,
+      price: currentPrice || BigInt(0),
```
mattcasey, line 346: not sure why, but there were some type errors not getting picked up in CI

PR #440, opened 2025-03-28
nft secondary marketplace listing

Devorein:


packages/scoutgame-ui/src/components/common/NFTListing/hooks/useGetDeveloperToken.ts
```diff
@@ -0,0 +1,15 @@
+import type { BuilderNftType } from '@charmverse/core/prisma-client';
+
+import { useGETImmutable } from '../../../../hooks/helpers';
+
+export function useGetDeveloperToken({ builderId, nftType }: { builderId: string; nftType: BuilderNftType }) {
```
mattcasey, line 5: please keep all hooks that call our /api into /hooks/api

packages/scoutgame/src/seaport/seaport.ts
```diff
@@ -0,0 +1,23 @@
...
+import type { Eip1193Provider } from 'ethers';
+import { JsonRpcProvider, BrowserProvider } from 'ethers';
+
+import { nftChain } from '../builderNfts/constants';
+
+export async function getSeaport() {
+  const chain = getChainById(nftChain.id);
+
+  if (!chain) {
+    throw new Error('Chain not found');
```
mattcasey, line 13: you should include the chain id in the error

packages/scoutgame/src/nftListing/createNftListing.ts
mattcasey, line 1: request: replace "create" with "record" and replace this (and other methods) to "recordNftListing" having createSeaportListing and createNftListing is confusing. this also follows teh pattern used for "recordNftMint"

packages/scoutgame/src/nftListing/cancelNftListingAction.ts
mattcasey, line 1: seems like this folder and seaport could go into a "marketplace" package to highlight that they belogn together, and keep scoutgame package small

packages/scoutgame/src/nftListing/purchaseNftListingAction.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const purchaseNftListingAction = authActionClient
+  .metadata({
+    actionName: 'purchase-nft-listing'
+  })
+  .schema(purchaseNftListingSchema)
+  .action(async ({ parsedInput, ctx }) => {
+    const scoutId = ctx.session.scoutId;
+
+    if (!scoutId) {
```
mattcasey, line 27: you're using the authActionClient, which already asserts scoutId is set

packages/scoutgame-ui/src/components/common/Card/BuilderCard/BuilderCard.tsx
```diff
@@ -27,9 +39,44 @@ export function BuilderCard({
...
   type: 'default' | 'starter_pack';
+  showListButton?: boolean;
+  scoutInView?: string;
 }) {
   const isDesktop = useMdScreen();
   const isLgScreen = useLgScreen();
+  const { user } = useUser();
+  const price = builder.price;
+
+  const userListings =
```
mattcasey, line 50: all this data and logic to determine a single value: lowestNonUserListing, should be done on the backend, only returning the single value to the frontend. It should just be written to the builderNft like we write currentPrice, so you don't have to calculate his on the fly all the time.

packages/scoutgame-ui/src/components/common/NFTListingPurchase/NFTListingPurchaseButton.tsx
mattcasey, line 1: request: rename these to "NFTListingDialog" and "NFTListingPurchaseDialog", to follow "NFTPurchaseDialog".. assuming all the ui is in dialogs

packages/scoutgame-ui/src/components/common/NFTListingPurchase/NFTListingPurchaseDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { NFTListingPurchaseForm } from './NFTListingPurchaseForm';
+
+type NFTListingPurchaseDialogProps = {
+  open: boolean;
+  onClose: VoidFunction;
+  listing: NFTListingPurchaseFormProps['listing'];
+  builder: NFTListingPurchaseFormProps['builder'];
+};
+
+function NFTListingPurchaseDialogComponent({ listing, open, onClose, builder }: NFTListingPurchaseDialogProps) {
```
mattcasey, line 18: it seems like this component could be generalized as a wrapper, so we dont have it repeated 3 times. its main intent is just to make sure u connect a wallet before u get to the form, right?

not a huge concern at the moment tho

mattcasey (changes requested):
see comments

packages/scoutgame/src/nftListing/cancelNftListingAction.ts
Devorein: its a bit tricky as there are some constants from scoutgame package that is used in those folders. I'm gonna leave it for now and get back to it in v2 when we have a good mockup in place

packages/scoutgame/src/nftListing/purchaseNftListingAction.ts
Devorein: oops good point

packages/scoutgame-ui/src/components/common/Card/BuilderCard/BuilderCard.tsx
Devorein: its a bit difficult to do it from the backend as sometimes we show the scouts listing and sometimes we show your listing sometimes we show the lowest listing. So for now I'm fetching them all and selecting the one to display on the fe. That is a bit easier but we can clean it up in v2

packages/scoutgame-ui/src/components/common/NFTListingPurchase/NFTListingPurchaseButton.tsx
Devorein: no unfortunately not all the UI is in a dialog, there is a button to trigger the dialog and a form inside the dialog so all of them are 3 different components. I went with `NFTListingDialog` initially but then realised that its not totally accurate

packages/scoutgame-ui/src/components/common/NFTListingPurchase/NFTListingPurchaseDialog.tsx
Devorein: lets do it for v2. we have 3 separate places you're right. nft purchase, nft listing and nft listing purchase.

PR #444, opened 2025-03-28
Weekly matchup

mattcasey:
For testing, i have been using a script in scripts/mockMatchupData.ts. It sends you some points and creates devs to pick from. I'd like to eventually update this to create github events, gems and ranks for the devs

All the major views have been implemented. Working features:
special callout on Scout page that shows top 3 scouts during the week
register for current week on Monday (subtracts 50 points)
you can register for next week after Monday (realized this is OK since we look at level, not rank) by visiting /matchup/register
the leaderboard page displays top scouts as well as current results and breakdown of your devs in the sidebar
mixpanel event for when someone registers

Not tested:
purchasing a new dev card during registration
display when there are multiple matchups
using production data

Not built:
OP airdrop/points payout at end of the week
notifications...?


.cursorrules
```diff
@@ -0,0 +1,39 @@
+You are an expert in TypeScript, Node.js, Next.js App Router, React, Material UI, Emotion CSS, and Prisma.
```
mattcasey, line 1: i noticed cursor's using interface and wanted to try adding rules. FWIW so far im not sure this is changing anything

packages/scoutgame/src/points/builderEvents/sendPointsForMiscEvent.ts
```diff
@@ -25,6 +25,20 @@ export async function sendPointsForMiscEvent({
   tx?: Prisma.TransactionClient;
 }) {
   const season = getCurrentSeason(week).start;
+  // make sure the sender has enough points
+  if (points < 0) {
```
mattcasey, line 29: added support to send negative points to take them away from a user. this way we don't need a new special builder event type

apps/scoutgame/components/common/EmblaCarousel/EmblaCarousel.tsx
mattcasey, line 1: Added a task card to remove swiper https://app.charmverse.io/scout-game/tech-task-list-14482155035143396?viewId=2d3dca88-ed4d-4b05-8a0f-57188d26a3aa&cardId=ff71af45-32fa-4833-b89f-6ad4eac52301

apps/scoutgame/components/common/EmblaCarousel/EmblaCarouselArrowButtons.tsx
mattcasey, line 1: the component files were copied from a generated example https://www.embla-carousel.com/examples/generator/

apps/scoutgame/components/common/Card/BuilderCard/BuilderCard.tsx
```diff
@@ -46,6 +50,9 @@ export function BuilderCard({
   type: 'default' | 'starter_pack';
   showListButton?: boolean;
   scoutInView?: string;
+  actionSlot?: ComponentType<T>;
```
mattcasey, line 53: I came up with this approach to customizing the action under each card, where we pass in a component and additional props for it. The requirements for adding a button to select a card included:
- be able to pass the week from server component into it
- but also receive the user id of the selected dev card, which is a client component

I used "any" to get typings to work, but if this becomes useful we can figure them out later.

apps/scoutgame/components/matchup/register/components/MatchupSelection/components/MatchupSelectionTabs.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useUser } from '@packages/scoutgame-ui/providers/UserProvider';
+import type { SyntheticEvent } from 'react';
+import { useState, Suspense } from 'react';
+
+interface MatchupSelectionGalleryProps {
+  myCardsView: React.ReactNode;
+  allCardsView: React.ReactNode;
+}
+
+export function MatchupSelectionTabs({ myCardsView, allCardsView }: MatchupSelectionGalleryProps) {
```
mattcasey, line 15: This tab component allows changing tabs that are rendered server-side without relying on the URL. I'm not sure if we have any other pattern for doing this, definitely open to a different approach.

PR #478, opened 2025-04-05
add dev card for matchup

mattcasey:
image


apps/scoutgame/components/matchup/register/components/MatchupSelection/components/MatchupSelectionTabsContext.tsx
```diff
@@ -0,0 +1,16 @@
+'use client';
+
+import { createContext, useContext } from 'react';
+
+type MatchupTabContextType = (tab: string) => void;
+
+export const MatchupTabContext = createContext<MatchupTabContextType | undefined>(undefined);
+
+// Hook to use the context
+export function useMatchupTab() {
```
mattcasey, line 10: i had to create a context for the tabs client to pass down the state so i could change tabs from a child, because there was a server component in-between.. cool that this works!

PR #494, opened 2025-04-09
add payout for matchup

mattcasey:


packages/blockchain/src/airdrop/deployThirdwebAirdropContract.ts
```diff
@@ -16,16 +18,16 @@ export async function deployThirdwebAirdropContract({
   totalAirdropAmount,
   expirationTimestamp,
   openClaimLimitPerWallet,
-  trustedForwarders,
-  proxyFactoryAddress,
-  implementationAddress,
+  trustedForwarders = [],
+  proxyFactoryAddress = THIRDWEB_AIRDROP_PROXY_FACTORY_ADDRESS,
```
mattcasey, line 22: i moved these constants down in the stack since it was just being imported in multiple places

packages/blockchain/src/airdrop/createThirdwebAirdropContract.ts
```diff
@@ -16,21 +16,15 @@ export async function createThirdwebAirdropContract({
...
   recipients: Recipient[];
   chainId: number;
   adminPrivateKey: `0x${string}`;
   tokenAddress: Address;
-  proxyFactoryAddress: Address;
-  tokenDecimals: number;
-  implementationAddress: Address;
   expirationTimestamp: bigint;
-  nullAddressAmount?: number;
+  nullAddressAmount?: string;
```
mattcasey, line 27: i changed this method a bit - we were already passing in values as stringified wei values, as i understand. but then we were converting to floats to add the total values up and then re-formatting them. so instead i just add numbers as BigInt which should be safer.

Devorein (approved):

PR #501, opened 2025-04-11
draft season UI

Devorein:
image

image

image

to test return `true` from `isDraftSeason`


apps/scoutgame/middleware.ts
```diff
@@ -1,17 +1,36 @@
...
 export async function middleware(request: NextRequest) {
   const session = await getSession();
   const isLoggedIn = !!session.scoutId;
   const path = request.nextUrl.pathname;
   const response = NextResponse.next(); // Create a response object to set cookies
   const platform = getPlatform();
+  const draftSeason = isDraftSeason();
+
+  if (draftSeason) {
+    if (!draftLinks.some((link) => path.startsWith(link))) {
```
mattcasey, line 20: i think this rule means no one can access claim pages, info pages, etc.

I don't think we really need any redirect rules here? "/" can be the landing page, and "/draft" can be the draft.

apps/scoutgame/middleware.ts
```diff
@@ -1,17 +1,36 @@
...
+      return NextResponse.redirect(new URL(isLoggedIn ? '/draft' : '/', request.url));
+    }
+
+    if (isLoggedIn) {
+      return NextResponse.redirect(new URL('/draft', request.url));
```
mattcasey, line 25: i tried running this locally and it just kept sending me to /draft even though i am already on it

packages/scoutgame-ui/src/components/common/Hidden.tsx
```diff
@@ -11,21 +11,21 @@ type Props = {
 } & BoxProps;
 
 // TODO: just use class names?
-export function getSXProps({ display = 'block', mdDown, mdUp }: Props) {
+export function getSXProps({ display = 'block', mdDown, mdUp, sx = {} }: Props) {
```
mattcasey, line 14: Does'nt make sense to pass 'sx' into a function that is just returning the value for 'sx', and its not efficient since u are creating an empty 'sx' prop every time it's called.

I prefer we just change Hidden to do: ```sx={{...getSXProps(), ...sx }}"

packages/scoutgame-ui/src/components/common/Hidden.tsx
```diff
@@ -11,21 +11,21 @@ type Props = {
...
+    return { display: { xs: display, md: 'none' }, ...sx };
   }
   log.warn('Hidden component must have either mdDown or mdUp prop');
-  return {};
+  return { ...sx };
 }
 
 // replace a deprecated Hidden component
-export function Hidden({ children, mdDown, mdUp, ...restProps }: PropsWithChildren<Props>) {
+export function Hidden({ children, display = 'block', sx = {}, mdDown, mdUp, ...restProps }: PropsWithChildren<Props>) {
```
mattcasey, line 26: you dont need to make an empty object for 'sx' even if it's optional

mattcasey:
On the landing page, there's too much space between the countdown and the action buttons. Try a maxWidth on the Container of 1024px, maybe.. i think you should also only use one Container for the whole page. that way, there is a consistent width (example: the rows in 'How to play' stick out further than the header)

mattcasey (changes requested):
left comments but particularly the middleware i don't think is necessary, and currently doesn't work for me. also this is a great candidate for e2e

apps/scoutgame/middleware.ts
Devorein: removed it. thnx

packages/scoutgame-ui/src/components/common/Hidden.tsx
Devorein: oops updated

PR #519, opened 2025-04-16
draft and airdrop frame

Devorein:


packages/scoutgame-ui/src/providers/AirdropAppProviders.tsx
mattcasey, line 1: this could probably live in the airdrop app, since nothing in scoutgame-ui is importing it and its specific to airdrop

mattcasey (approved):

packages/scoutgame-ui/src/providers/AirdropAppProviders.tsx
Devorein: good point

PR #509, opened 2025-04-14
draft v2

Devorein:

PR #520, opened 2025-04-17
draft v3

Devorein:
1. redirect user to /draft page after onboarding
2. format bid amount (to remove trailing 0s)
3. use drafts logger (previously was using mints logger)
4. use v4 box api for draft useBoxAction (forgot to replace it in #509)
5. redirect from /accounts page
6. Bid button takes you to login page (when unauthed)
7. direct transfer of DEV token
8. check if the address is already in use (just like nft purchase)
9. draft bid transaction failed notification + email


mattcasey (approved):

apps/scoutgame/components/common/DraftDeveloperInfoModal/components/DraftDeveloperBidForm.tsx
```diff
@@ -72,11 +75,13 @@ function DraftDeveloperBidFormComponent({
 
   // Fetch ETH and LINK prices from CoinGecko
   const { data: prices, isLoading: isLoadingPrices } = useSWR('token-prices', async () => {
-    const response = await fetch('https://api.coingecko.com/api/v3/simple/price?ids=ethereum,talent&vs_currencies=usd');
+    const response = await fetch(
```
mattcasey, line 78: we still need to replace coingecko with decent's response right?

mattcasey (approved):
so many changes!

apps/scoutgame/components/common/DraftDeveloperInfoModal/components/DraftDeveloperBidForm.tsx
Devorein: correct but decent is not giving the exchange rate rn. Its 0

PR #522, opened 2025-04-17
send airdrop live email to users

Devorein:
[Card](https://app.charmverse.io/scout-game/airdrop-email-april-21-7170509382669856)


scripts/sendAirdropLiveEmail.ts
```diff
@@ -0,0 +1,44 @@
+import { prisma } from '@charmverse/core/prisma-client';
+import mailgunClient, { DOMAIN } from '@packages/mailer/mailgunClient';
+
+export async function sendAirdropLiveEmail() {
+  const users = await prisma.scout.findMany({
+    where: {
+      email: { not: null },
```
mattcasey, line 7: i think we should be more picky about who gets the emails...
- hasnt been deleted
- maybe has verified their email (but we'd probably want to make an exception for people that actually received an airdrop)

mattcasey (changes requested):
see comment

scripts/sendAirdropLiveEmail.ts
Devorein: we would need the airdrop list for email verification. I will merge it for now and since its a one time thing deal with it when we send the emails

PR #541, opened 2025-04-22
Remove references to special "onchain" env

mattcasey:
Searched and removed the use of "isOnchainPlatform()", "onchain_webapp" and "onchain_cron", as well as "builderTokenDecimals" except for some unused methods related to points

It's a WIP because i want to merge the test fix PR first, and then see what tests may be failing here


apps/scoutgame/components/matchup/register/components/MatchupSelection/components/AllDevelopersTable/AllDevelopersTable.tsx
```diff
@@ -215,7 +215,7 @@ export function AllDevelopersTable({
                   builder={{
                     ...builder,
                     builderStatus: 'approved',
-                    price: BigInt(2 * 10 ** builderTokenDecimals)
+                    price: BigInt(2 * 10 ** devTokenDecimals)
```
Devorein, line 218: is this correct? This should be 100 DEV right?

Devorein (changes requested):

PR #561, opened 2025-04-25
onchain v1

Devorein:


apps/cron/src/tasks/processGemsPayout/__tests__/processScoutPointsPayout.spec.ts
mattcasey, line 1: most of these tests look like they are worth keeping around but with the new onchain method

apps/cron/src/tasks/processOnchainGemsPayout/SablierMerkleFactory.json
mattcasey, line 1: maybe we can just delete this

apps/cron/src/worker.ts
```diff
@@ -61,7 +60,7 @@ addTask('/approve-developers', approveDevelopers, approveDevelopersLog);
 
 addTask('/send-push-notifications', sendNotifications);
 
-addTask('/process-gems-payout', processGemsPayout, processGemsPayoutLog);
+addTask('/process-gems-payout', processOnchainGemsPayout, processOnchainGemsPayoutLog);
```
mattcasey, line 63: remember to update cron.yml in the onchain-cron app

packages/scoutgame/src/protocol/generateWeeklyClaims.ts
```diff
@@ -64,16 +64,15 @@ export async function generateWeeklyClaims({
   // get end of the next season, or this one if there are no more
   const nextSeason = getNextSeason(getCurrentSeasonStart()) || getCurrentSeasonStart();
   const weeksPerSeason = getSeasonConfig(nextSeason).weeksPerSeason;
-  const validUntil = getDateFromISOWeek(nextSeason).plus({ weeks: weeksPerSeason }).toISO();
+  const validUntilDate = getDateFromISOWeek(nextSeason).plus({ weeks: weeksPerSeason });
+  const validUntil = Math.floor(validUntilDate.toSeconds());
```
mattcasey, line 68: was this just a bug? why did we switch from the isodate to unix time?

packages/scoutgame/src/builders/getDevelopersForTable.ts
```diff
@@ -379,7 +362,12 @@ export async function getDevelopersForTable({
     const lastItem = builderNfts[builderNfts.length - 1];
     const nextCursor =
       lastItem && developers.length === limit
-        ? { id: lastItem.createdAt.toISOString(), value: Number(lastItem.currentPrice), order, sortType: sortBy }
+        ? {
+            id: lastItem.createdAt.toISOString(),
+            value: Number(lastItem.currentPriceDevToken),
```
mattcasey, line 367: this should be currentPrice since that's what we are sorting by

apps/scoutgame/components/matchup/components/RegistrationButton.tsx
```diff
@@ -73,7 +73,13 @@ export function RegistrationButton({ registered, week }: { registered: boolean;
         disabled={registered}
         variant='contained'
         color='secondary'
-        endIcon={registered ? <CheckCircleIcon color='inherit' /> : <PointsIcon color='inherit' />}
+        endIcon={
+          registered ? (
+            <CheckCircleIcon color='inherit' />
+          ) : (
+            <Image src='/images/dev-token-logo.png' alt='DEV token' width={18} height={18} />
```
mattcasey, line 80: TODO later: we should make this a component so we're not hardcoding the image path everywhere

packages/dates/src/config.ts
```diff
@@ -6,6 +7,7 @@ export type ISOWeek = string; // isoweek, e.g. '2024-W01'
 
 export type SeasonConfig = {
   title: string;
+  id: string;
```
mattcasey, line 10: MUST FIX: we already use the 'start' as the pointer to seasons. this jsut introduces a brand new way to reference seasons and i dont think that is helpful

PR #595, opened 2025-05-01
Onchain token claims

Devorein:


apps/cron/src/notifications/sendGemsPayoutNotifications.ts
```diff
@@ -120,14 +120,14 @@ export async function sendGemsPayoutNotifications({ week }: { week: string }) {
...
-      if (weeklyClaimablePoints) {
+      const weeklyClaimableTokens = await getClaimableTokens({ userId: scout.id, week });
+      if (weeklyClaimableTokens) {
         await sendNotifications({
           userId: scout.id,
           email: {
             templateVariables: {
               name: scout.displayName,
-              points: weeklyClaimablePoints,
+              points: weeklyClaimableTokens,
```
mattcasey, line 130: do we need to change the mailgun templates?

packages/scoutgame/src/tokens/checkIsProcessingPayouts.ts
mattcasey, line 1: Nice-to-have: i think we could potentially just change this to look for the actual airdrop, then we wouldn't need the weeklyClaims table at all?

packages/scoutgame/src/tokens/getClaimableTokens.ts
```diff
@@ -41,10 +43,10 @@ export async function getClaimablePoints({
     }
   });
 
-  const totalUnclaimedPoints = pointsReceipts.reduce((acc, receipt) => acc + receipt.value, 0);
+  const totalUnclaimedTokens = tokensReceipts.reduce(
+    (acc, receipt) => acc + Number(formatUnits(BigInt(receipt.value), 18)),
```
mattcasey, line 47: nice-to-have: its best to import devTokenDecimals instead of hard-coding 18... its a nice reference also that these tokens are dev tokens.

packages/scoutgame/src/tokens/divideTokensBetweenDeveloperAndHolders.ts
```diff
@@ -85,49 +78,49 @@ export function divideTokensBetweenBuilderAndHolders({
...
-      scoutsRewardPool: earnableScoutTokens
+      scoutsRewardPool: earnableTokens
     });
     const scoutTokens = Math.floor(scoutReward);
 
     return { scoutId: owner.scoutId, nftTokens: owner.totalNft, erc20Tokens: scoutTokens };
   });
 
-  const tokensForBuilder = Math.floor((defaultBuilderPool * earnableScoutTokens) / 100);
+  const tokensForDeveloper = Math.floor((defaultDeveloperPool * earnableTokens) / 100);
```
mattcasey, line 88: all the division done should probably be done with BigInts. This removes the need to use Math.floor, and allows more accuracy

packages/scoutgame/src/partnerRewards/getPartnerRewardsForScout.ts
```diff
@@ -157,7 +154,7 @@ export async function getPartnerRewards({
 
   partnerRewardPayouts.forEach((payout) => {
     const partnerReward: PartnerRewardBase<any> = {
-      points: Number(formatUnits(BigInt(payout.amount), payout.payoutContract.tokenDecimals)),
+      tokens: Number(formatUnits(BigInt(payout.amount), payout.payoutContract.tokenDecimals)),
```
mattcasey, line 157: are we handling it properly when you have a decimal payout? (ie maybe calling toFixed() in the frontend)?

apps/scoutgame/components/claim/components/common/TokensCell.tsx
```diff
@@ -1,10 +1,10 @@
 import { Stack, Typography } from '@mui/material';
 import Image from 'next/image';
 
-export function PointsCell({ points }: { points: number }) {
+export function TokensCell({ tokens }: { tokens: number }) {
   return (
     <Stack direction='row' alignItems='center' justifyContent='flex-end' gap={0.5}>
-      <Typography>{points}</Typography>
+      <Typography>{tokens}</Typography>
```
mattcasey, line 7: pretty sure we should format this number using, eg, ceilToPrecision(tokens, 4).toLocaleString()

packages/dates/src/config.ts
```diff
@@ -60,15 +64,17 @@ export const seasons: SeasonConfig[] = [
...
   {
     start: '2025-W18', // April 28th 2025
     title: 'Season 1',
-    starterNftAddress: '0x77ef845f8b2b7b40b68af10d1031313983ccf5a2' as Address,
-    standardNftAddress: '0x1aa94658c5586284bb7815e590a3456f76901500' as Address,
-    weeksPerSeason: 13
+    starterNftAddress: '0x9b11a12f267b21580ef911e404e96659d27eef84' as Address,
+    standardNftAddress: '0xa32f8737513454d6a938359614fdf47838a2b6d7' as Address,
+    weeksPerSeason: 13,
+    allocatedTokens: 2500
```
mattcasey, line 77: shouldn't this be 	1,200,000? https://scoutgame.xyz/info/dev-tokens

apps/scoutgame/app/api/session/claimable-tokens/route.ts
```diff
@@ -0,0 +1,22 @@
...
+  processingPayouts: boolean;
+};
+
+export async function GET() {
+  const user = await getUserFromSession();
+  if (!user) {
+    return NextResponse.json({ tokens: 0 });
+  }
+
+  const processingPayouts = await checkIsProcessingPayouts({ week: getLastWeek() });
```
mattcasey, line 18: Checking this seems unecessary, aren't you looking for claimable tokens for all weeks?

I think the reason we needed processingPayouts before was because we would write pointsReceipts as we processed dev activity

apps/cron/src/notifications/sendGemsPayoutNotifications.ts
Devorein: not necessary since chris already did that

packages/scoutgame/src/protocol/calculateWeeklyClaims.ts
```diff
@@ -114,68 +111,128 @@ export async function calculateWeeklyClaims({
...
   const allClaims = await Promise.all(
-    topWeeklyBuilders
+    topWeeklyDevelopers
       // We only want to issue claims for builders that have sold at least one NFT
-      .filter(
-        (builder) =>
-          !!tokenBalances[builderNfts.find((nft) => nft.builderId === builder.builder.id)!.tokenId.toString()]
-      )
+      .filter((developer) => {
+        const starterNft = builderNfts.find(
```
mattcasey, line 142: Two questions:
- Don't we need to still check that there is a token balance for their NFTs? Otherwise this is just going to include all developers.
- Can you just use nftType to determine starter vs default? This is better as it doesn't rely looking up the season configuration.

packages/dates/src/config.ts
Devorein: all of these needs (the contracts and allocated tokens) to be changed. I used 2500 for testing purposes. I'm gonna have a different config for onchain

apps/scoutgame/app/api/session/claimable-tokens/route.ts
Devorein: No this just checks if the airdrop was deployed by our protocol. i.e as soon as our protocol sets the weekly merkle tree root we have a weekly claim ready for the week.

PR #603, opened 2025-05-05
update all time stats when we update season stats for tokens earned

mattcasey:


apps/scoutgame/lib/actions/handleOnchainClaimAction.ts
```diff
@@ -92,10 +95,33 @@ export const handleOnchainClaimAction = authActionClient
...
-          increment: Number(formatUnits(pointsEarnedAsDeveloper, devTokenDecimals))
+          increment: pointsEarnedAsDeveloperNumber
+        },
+        pointsEarnedAsScout: {
+          increment: pointsEarnedAsScoutNumber
+        }
+      }
+    });
+
+    await prisma.userAllTimeStats.upsert({
```
mattcasey, line 106: we need to also update all time stats

PR #606, opened 2025-05-07
developer api route

Devorein:


apps/scoutgame/app/api/developer/route.ts
```diff
@@ -0,0 +1,80 @@
...
+        select: {
+          week: true,
+          gemsCollected: true
+        }
+      }
+    }
+  });
+
+  if (!developer) {
+    return NextResponse.json({ developer: null }, { status: 200 });
```
mattcasey, line 65: should return a 400

mattcasey (approved):

apps/scoutgame/app/api/developer/route.ts
Devorein: sure

PR #610, opened 2025-05-07
Show a warning on matchup registration if you do not own enough nfts

mattcasey:
image

there was some unrelated cleanup in this PR too


packages/scoutgame/src/tokens/__tests__/calculateDeveloperLevel.spec.ts
Devorein, line 1: Thank you for fixing this. It was flakey recently.

Devorein (approved):

PR #613, opened 2025-05-08
fix method for token ownership

mattcasey:


packages/scoutgame/src/protocol/resolveTokenOwnership.ts
```diff
@@ -69,18 +72,22 @@ export async function resolveTokenOwnership({
   week: ISOWeek;
   chainId: number;
 }): Promise<TokenOwnership> {
-  const lastBlock = await getLastBlockOfWeek({ week, chainId });
-  const standardNftContractAddress = getNFTContractAddress(week);
-  const starterNftContractAddress = getStarterNFTContractAddress(week);
+  const toBlock = await getLastBlockOfWeek({ week, chainId });
+  const season = getCurrentSeasonStart(week);
```
mattcasey, line 76: there was actually a critical bug, we were getting nft contract by week, but it throws an error if it is not provided a season

packages/scoutgame/src/protocol/getNftsByOwner.ts
```diff
@@ -0,0 +1,100 @@
+import { getLastBlockOfWeek } from '@packages/blockchain/getLastBlockOfWeek';
```
mattcasey, line 1: we are not using this method anywhere except my script but it might be useful in the future

packages/dates/src/config.ts
```diff
@@ -78,6 +80,7 @@ export const seasons: SeasonConfig[] = [
     title: 'Season 1',
     starterNftAddress: '0x77ef845f8b2b7b40b68af10d1031313983ccf5a2' as Address,
     standardNftAddress: '0x1aa94658c5586284bb7815e590a3456f76901500' as Address,
+    nftBlockNumber: 29_504_000,
```
mattcasey, line 83: track this here to optimize the token owner lookup

PR #637, opened 2025-06-05
link issues and support good dollar airdrop payout

Devorein:


mattcasey (approved):
- suggest wrapping the logic to determine payout in a new method that we can reuse in admin and low balance alert
- include a column for issue tags when we export the PRs in admin

PR #657, opened 2025-06-23
revive developer nft listing

Devorein:


packages/scoutgame/src/nftListing/recordNftListing.ts
```diff
@@ -45,17 +47,30 @@ export async function recordNftListing({
     }
   });
 
-  // Create the listing in our database
+  const builderNft = await prisma.builderNft.findUniqueOrThrow({
```
mattcasey-fountain, line 50: this doesn't look like it is being used?

packages/scoutgame/src/nftListing/updateCurrentNftListingPrice.ts
```diff
@@ -0,0 +1,45 @@
...
+          priceDevToken: true
+        }
+      }
+    }
+  });
+
+  let lowestListingPrice: bigint | null = null;
+
+  if (builderNft.currentListingPrice) {
+    lowestListingPrice = BigInt(builderNft.currentListingPrice);
```
mattcasey-fountain, line 24: currentListingPrice is already a BigInt, i guess it shouldnt matter to use the constructor again tho

packages/scoutgame/src/nftListing/updateCurrentNftListingPrice.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  // Find the lowest price among active listings
+  for (const listing of builderNft.listings) {
```
mattcasey-fountain, line 28: suggestion:
```
const lowestPrice = Math.min(listings.map(listing => listing.priceDevToken));
```

mattcasey-fountain (approved):

packages/scoutgame/src/nftListing/updateCurrentNftListingPrice.ts
Devorein: math min doesn't work with bigint

